PR #4, opened 2016-10-24
API: let the modified time be changeable

korelstar:
It will close owncloud/notes#247:

> If a note is created/edited using a 3rd-party app, this can be done in offline mode. Therefore, there can be a time difference between editing the note on the client and putting the changed data to the server through the API. However, the modified time will be set to the time when putting the data to the server.
> 
> This behavior can be confusing for the user, since the shown modified time does not reflect what (when) the user did.
> 
> I propose to extend the API and introduce a new optional field `modified` which can be passed when creating or updating a note, beside `content`.
### Tasks
- [x] API: let the modified time be changeable
- [x] API: allow setting favorites in `create`, refactoring


Henni:
Note to myself: this also adds favorite functionality

controller/notesapicontroller.php
```diff
@@ -125,23 +127,35 @@ public function create($content) {
      *
      * @param int $id
      * @param string $content
+     * @param int $modified
      * @param boolean $favorite
      * @return DataResponse
      */
-    public function update($id, $content=null, $favorite=null) {
+    public function update($id, $content=null, $modified=0, $favorite=null) {
```
Henni: What happens if this defaults to `0`?

korelstar: If `modified` is not stated in the request, the behavior is the same as before of my changes: the `mtime` of the note's file is set to the current date/time.

Therefore, my changes of the API are totally backward compatible. However, the API documentation has to be revised. But the wiki has not been transferred, yet.

Henni: :+1:

korelstar:
I'm done -- from my site, this can be merged.

Henni:
Looks good to me.
@stefan-niedermann would you like to take a look, as this affects your awesome app.

stefan-niedermann:
Yeap, i have already subscribed this issue :) Will have a look at implementation when i have some free time

korelstar:
Can this be merged, now?

Henni:
@stefan-niedermann ping

stefan-niedermann:
LGTM 👍 
Ready to merge from my point of view :smiley:

Henni:
Awesome! Happy we got this in! :tada:

Issue #2, opened 2016-10-21
Put star on the left side of the note


jancborchardt:
Currently it’s on the right side. We should put it on the left side to be consistent with the sidebars in other apps.

For that, the `<ul>` just needs the `class="with-icon"` to have padding to the left. Then the star can be positioned on the left of it.

cc @korelstar @Henni

jancborchardt:
(Then it’s also not that close to the delete button anymore which is a bit safer. ;)

korelstar:
@jancborchardt Hmmm. Typically, the icon left to the text in the navigation sidebar is a visual representation of the text/entry and hence represents the navigation entry without providing a new action. This applies to `calendar`, `files` and `tasks` (I'm not aware of another usage!).

For `notes`, the icon has it's own action which is not choosing the navigation entry like in the other apps. So it's a complete other use of the icon. If both the icon are at the same position in both use cases, this could be confusing.

jancborchardt:
@korelstar good points. However even for Calendar, the navigation is very different. There, clicking the navigation entry as well as circle toggles visibility instead of switching to a different group.

And in Notes the navigation is a file list. For Files, the file list has the star function on the left of the file. That’s also where it would belong for Notes, even though the list is in the sidebar. :) It’s definitely strange next to the delete button, having the action of »I like this note a lot« and »I hate this note and it’s useless« so close to each other. ;)

iSpeeX:
"For notes, the icon has it's own action"
I didn't understand your answer about that left-icon in Notes. In my Notes app (web), i have no icons on any note, is it normal ?

korelstar:
Maybe, you just haven't any notes marked as favorite. If you mouseover a note in the navigation bar, you see a gray star (currently on the right from the note's title). If you click on that star, the note will be marked as favorite.

However, in version 3.0 of the notes app, this behavior will change (see #290 and the discussion under nextcloud/nextcloud-vue#246) and we will have the star icon on the left side as requested by @jancborchardt

iSpeeX:
Well I have some favorites but still no icons on the left of notes titles. I think i may have misunderstand the whole point.

Thank you for you answer, you guys seems to be very comitted to this project, it's really nice

korelstar:
[...]

PR #6, opened 2016-10-24
Also show notes inside subdirectories

Henni:
First step towards owncloud/notes#2

TODO:
- [x] Fix (Maybe rewrite) tests


Henni:
Seems like a transifex update from owncloud snuck in. I'll "force-fix" this.

korelstar:
Nice! I'm really looking forward to this feature.

Do you plan a "silent" integration of subdirs (the source of the note is omitted, they are all shown in one list of notes) or should the subdirs be shown in the GUI as "folders" (like a contact group in the contacts app)? The second variant would directly influence #2, because the navigation sidebar could then show the list of folders instead the list of notes and a second sidebar could be needed (like in the contacts app). I like that concept because it would be possible to structure the notes in a better way.

/cc @jancborchardt

Henni:
For not it is "silent".
The goal is to create some kind of structure (categorization) with the help of the subdirectories.
But I'd prefer to keep it simple and maybe integrate it into the already existing sidebar. (let's talk about this in #8)

jancborchardt:
CC'ing the people who were involved in the discussion on subfolders for Notes in the original issue. :) You’ll probably like that work is being done on it here! :tada:

Reviews are very welcome!
@caco3 @pyluyten @lpalgarvio @reimersjan @slashrsm @Jas2Ma @grefel @pierrejochem @greenais @rakshazi @sprnza @pbek @olantrust @xraMsamohT @dasmaetthes @jibweb

dasmaetthes:
Very nice,
Would it work also with owncloud or nextcloud only?

jancborchardt:
@dasmaetthes it will likely work with ownCloud as well for the near future. However as most people developing on these apps (also like Calendar, Contacts and Mail) moved to Nextcloud and don’t necessarily test with both systems, bugs could occur in the future. So we recommend to upgrade. :)

Henni:
@dasmaetthes *should* work with owncloud as well. Tests on owncloud are welcome :)

dasmaetthes:
Seems to work well with owncloud.

Thank you very much.

evaryont:
Just wondering if there is anything blocking this from being merged?

Henni:
@evaryont the tests should be fixed before merging this.

korelstar:
@Henni 
I've fixed the tests, rebased the branch and added a new attribute `category` for the `Note` entity which is filled with the subdir name. I suggest to merge this now since some people are waiting for this impatiently ... :grin:

Henni:
I removed two debug lines which I left in.

After travis has finished I'll merge this and release notes 2.2

jancborchardt:
Awesome work! :) :tada:

PR #12, opened 2016-10-25
Add distraction free mode

### LINKED ISSUE #9: simplenote like UI
distraction free writing (app and webfrontend)

Henni:
fixes #9 

cc @irgendwie @dk87 @jancborchardt


jancborchardt:
Awesome! This is very cool. One enhancement to make it really distraction-free and true to the icon would be to actually use the browser fullscreen API.

irgendwie:
Maybe add a tool-tip, describing the function of the new button.
Additionally, the button (and word-count) is barely visible - increase contrast? :blush:

Henni:
@jancborchardt I'm planning to do that, it is a bit more difficult than I thought because it is still prefixed in all browsers.

karlitschek:
indeed super cool feature!

Henni:
@irgendwie does this look better to you?

jancborchardt:
@Henni this is how we do it in the Spreed app: https://github.com/nextcloud/spreed/commit/c606751fb2f5ca17b6d3f0525b1f47e3b16e3d85 :)

irgendwie:
@Henni Contrast is much better! Thanks!
The tooltip needs to be placed a bit higher I think and should not be labeld "test" :wink:

jancborchardt:
@Henni what do we have left to do here btw? :D

jancborchardt:
@Henni @irgendwie @korelstar can we merge this? :)

Henni:
Final review and then we are ready to go!

korelstar:
1. Travis failed (but this may has nothing to do with these changes)
2. These changes are not compatible with the up-to-date stable Nextcloud 10.0.1 (and below). See nextcloud/server#1818. Is it possible to improve this situation?
3. When distraction free mode is activated, I see black font on black background. The only thing, I can see is the red underlining of the spell correction. I'm using Firefox ESR 45.5.1 on Debian jessie (but this could also be due to other changes in core that are not backported yet to Nextcloud 10 -- I don't know, I only tested with Nextcloud 10.0.1). If I add `#app-content-container { background: #fff; }` to `notes.css` it works.

Henni:
1. I'll look into it
2. Do I understand it correctly, that versions before 10.0.1 didn't have this icon? If this is the case it might be possible to find a fallback, which could be removed in the future
3. Black on black is really strange... No idea where the black background comes from. (Also can't reproduce it)

korelstar:
2. No, in fact Nextcloud 10.0.1 does *not* have this icon. This will be only part of version 11 and 10.0.2 (if I understood the release strategy correctly).
3. Black seems to be the default background color for fullscreen in Firefox. I could reproduce this on another computer with Firefox 50.0.2 on Windows.

I tried it also with this minimal page ("works" only on Firefox):
```
<html>
<body>
<div id="content">
click me
</div>
<script>
function toggleFullScreen() {
        if(document.mozFullScreenElement) {
                document.mozCancelFullScreen();
        } else {
                document.getElementById('content').mozRequestFullScreen();
        }
}
document.getElementById("content").addEventListener("click", toggleFullScreen, false);
</script>
</body>
</html>
```

Henni:
@korelstar I should be able to fix both of these points. I'll take a look

jancborchardt:
Yeah, so since the fullscreen-icon is only going to be in the latest patch, probably best to copy it over into the img folder, and add a class `.icon-fullscreen` to the Notes app CSS which calls that one. That will be the fallback for older systems.

Henni:
fixed the black background in firefox.

@korelstar @jancborchardt regarding the second issue: is it still worth supporting Nextcloud 9 (and 10)?

PR #26, opened 2016-12-08
Update README.md

lekkerduidelijk:
Update owncloud paths to nextcloud. I noticed this when installing the app again after migrating to nextcloud.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,9 @@ For further developer and user documentation please visit [the wiki](https://git
 ##### Minimum PHP Version
 * PHP >= 5.4
 
-##### Minimum Nextcloud Version
-* >= 9.0
+##### Minimum ownCloud / Nextcloud Version
+* ownCloud >= 8.1
+* Nextcloud >= 9.0
 
 ##### Supported Webservers
 * Apache
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ For further developer and user documentation please visit [the wiki](https://git
 * PHP >= 5.4
 
 ##### Minimum Nextcloud Version
-* >= 8.1
+* >= 9.0
 
 ##### Supported Webservers
 * Apache
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # Notes
 
-[![build state](https://travis-ci.org/owncloud/notes.png)](https://travis-ci.org/nextcloud/notes)[![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/nextcloud/notes/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/nextcloud/notes/?branch=master)
+[![build state](https://travis-ci.org/nextcloud/notes.png)](https://travis-ci.org/nextcloud/notes)[![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/nextcloud/notes/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/nextcloud/notes/?branch=master)
 
 The Notes app is a distraction free notes taking app. It offers a [RESTful API](https://github.com/nextcloud/notes/wiki/API-0.2) for app developers. The source code is [available on GitHub](https://github.com/nextcloud/notes)
 
-![Screenshot of ownCloud Notes](https://cloud.githubusercontent.com/assets/4741199/17731273/e557b5fe-646c-11e6-9975-d6b242454482.png)
+![Screenshot of Nextcloud Notes](https://cloud.githubusercontent.com/assets/4741199/21027342/b70a6be2-bd90-11e6-9f12-eca46d6c505a.png)
 
 For further developer and user documentation please visit [the wiki](https://github.com/nextcloud/notes/wiki)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,12 +1,12 @@
 # Notes
 
-[![build state](https://travis-ci.org/owncloud/notes.png)](https://travis-ci.org/owncloud/notes)[![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/owncloud/notes/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/owncloud/notes/?branch=master)
+[![build state](https://travis-ci.org/owncloud/notes.png)](https://travis-ci.org/nextcloud/notes)[![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/nextcloud/notes/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/nextcloud/notes/?branch=master)
 
-The Notes app is a distraction free notes taking app. It offers a [RESTful API](https://github.com/owncloud/notes/wiki/API-0.2) for app developers. The source code is [available on GitHub](https://github.com/owncloud/notes)
+The Notes app is a distraction free notes taking app. It offers a [RESTful API](https://github.com/nextcloud/notes/wiki/API-0.2) for app developers. The source code is [available on GitHub](https://github.com/nextcloud/notes)
 
 ![Screenshot of ownCloud Notes](https://cloud.githubusercontent.com/assets/4741199/17731273/e557b5fe-646c-11e6-9975-d6b242454482.png)
```
MariusBluem: Could you take a new screenshot? 🙈

README.md
```diff
@@ -1,12 +1,12 @@
...
 
-For further developer and user documentation please visit [the wiki](https://github.com/owncloud/notes/wiki)
+For further developer and user documentation please visit [the wiki](https://github.com/nextcloud/notes/wiki)
```
MariusBluem: Remove this line until we have a wiki... It is currently disabled 😁

README.md
korelstar: The screenshot was taken by @stefan-niedermann, maybe he has still the original (plain) text (and list of notes) and could create a new one.

README.md
```diff
@@ -1,12 +1,12 @@
 # Notes
 
-[![build state](https://travis-ci.org/owncloud/notes.png)](https://travis-ci.org/owncloud/notes)[![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/owncloud/notes/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/owncloud/notes/?branch=master)
+[![build state](https://travis-ci.org/owncloud/notes.png)](https://travis-ci.org/nextcloud/notes)[![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/nextcloud/notes/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/nextcloud/notes/?branch=master)
```
korelstar: Here is still one reference to owncloud. Can you please change this, too?

README.md
Henni: Reactivating would make sense.
We should mention the API and all the third party clients, like @stefan-niedermann's
I believe we should be able to transfer it from owncloud. I'll have a look.

lekkerduidelijk:
I've made some changes. Let me know if this is ok.

README.md
```diff
@@ -16,7 +16,7 @@ For further developer and user documentation please visit [the wiki](https://git
 ##### Minimum PHP Version
 * PHP >= 5.4
 
-##### Minimum ownCloud Version
+##### Minimum Nextcloud Version
 * >= 8.1
```
Henni: I'm not completely sure. But wasn't 9.0 the first version of Nextcloud?
We should replace this version number accordingly.

korelstar: Is owncloud now officially unsupported by this app? Another way would be to list both Nextcloud and owncloud.

Henni: We should list both. So something like owncloud >= 8.1 and nextcloud >= 9.0

lekkerduidelijk: Think you are right. Wait up...

lekkerduidelijk: I've listed both for now. I can imagine Nextcloud Notes turning into a separate codebase with the upcoming Nextcloud version 11.

Henni:
👍 
@MariusBluem are you also happy with the latest changes?
Thumbs up from you and I think we can merge this.

README.md
```diff
@@ -16,8 +16,9 @@ For further developer and user documentation please visit [the wiki](https://git
 ##### Minimum PHP Version
 * PHP >= 5.4
 
-##### Minimum ownCloud Version
-* >= 8.1
+##### Minimum ownCloud / Nextcloud Version
```
MariusBluem: Don't we want to remove ownCloud completely as we have done it in the other apps?

stefan-niedermann: I agree. If it runs - fine. But otherwise it could mean there is complete official support for ownCloud (and i don't think there is, is? ;-) )

Henni: The code of ownCloud and Nextcloud hasn't diverged in any way, that affects the notes app, yet.
So I would like to still keep this information in the README.
But we might drop support for ownCloud in the future, if something significant changes. We can always say "ownCloud version 8.1 - ??.??". But for now it works and I'm happy to support it.

MariusBluem:
Much better than before ... We can overthink this later 😉 THX!

PR #28, opened 2016-12-10
Simple search functionality

### LINKED ISSUE #25: Ability to search notes
Hi dev, it is nice to have a search function for notes. I have plenty of notes, which is no problem searching in android. But the problem is when I use Internet Browser from Windows I have to search all the notes, it kinda dizzy.. :)

Henni:
fixes #25 

looking forward to your reviews
@affanse @korelstar @phsc @putt1ck @ThomasDaheim @jancborchardt


Henni:
@korelstar Am I right when I guess that you are using a Firefox version <= 48? That would explain the problem with the icon positioning (http://caniuse.com/#feat=background-position-x-y). If this is the case, I believe I know how to fix this.
I'm not sure why the text isn't selectable but I'll experiment and try to figure it out.

enhancement: noted - could be fixed in a separate PR, but shouldn't be a huge deal (has been done before http://stackoverflow.com/a/23506051)

I have created an in-app search, because I have heard a rumor that @jancborchardt is unhappy with the global search 😉

korelstar:
Yes, I'm using the current ESR version of Firefox and that's 45.5.1.

I've found out, why the text isn't selectable: because the `<input>` is inside of an `<a>` tag. Could you remove the `<a>` tag and use another method to show the search icon?

Henni:
@korelstar possible but requires duplicating some styles from core. But you are right, that is probably better than wrong markup.

korelstar:
Maybe you can use a `<div>` or `<span>` instead of the `<a>`. That seems to work.

phsc:
Thank you for this feature! Works great on Chrome 54.
Regarding the multi word search, I agree with @korelstar. It would be a very nice enhancement, if the search would find all notes containing wordA and wordB at any position.

Henni:
@korelstar @phsc I implemented the changes to the search functionality you requested.

korelstar:
:+1: Great work!

phsc:
Wonderful! Just tested it on Chrome 54 and Nextcloud 10.0.1. Works like a charm.
- No problem with german umlauts.
- No problem with Markdown formatting codes in between.

Ready to be merged, from my point of view. 👍

PR #34, opened 2016-12-14
Add to info.xml and automate appstore release

Henni:
preparing 2.1.1


Henni:
@korelstar please take another look

PR #52, opened 2017-01-20
[Help needed] Travis: debug CI-testing on Nextcloud

korelstar:
The error found by @Henni using Drone in the integration test (see #36) also appears with Travis when I wanted to switch the test to Nextcloud. Therefore, I did some investigation. This Pull-Request is only for testing purpose. I will go into the details later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,8 @@ script:
 
 matrix:
   include:
+    - php: 7.0
+      env: "DB=mysql CORE_BRANCH=stable12"
     - php: 5.6
       env: "DB=mysql CORE_BRANCH=stable11"
     - php: 5.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 php:
   - 5.6
-  - 7.0
 
 env:
   global:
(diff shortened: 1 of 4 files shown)


korelstar:
Okay, I've lost very much time for this today. Now, I'm asking for help.

I was able to isolate the error source why our tests are failing for `stable11` branch of Nextcloud/server while they are fine for `stable10`. My debugging output show that for `stable11`, newly created files are included in `Folder::getDirectoryListing()` with the correct ID, but `Folder::getById()` nevertheless returns an empty result. However, for `stable10`, `Folder::getById()` returns the correct file, too.

I've found out, that @icewind1991 did some optimization work on `Folder::getById()` (see nextcloud/server#1441, which is the only difference between `stable10` and `stable11` in this method). Therefore, I'm considering that a correlation may be possible :sunglasses: . At least @icewind1991 is someone who perfectly knows this part of the server code, and maybe could help us in our beautiless situation. Could you please have a look on this, @icewind1991 ?  :relaxed:

korelstar:
It looks like the problem is, that the `UserMountCache` is empty for our tests. This means, that we have to add an appropriate `mount` for our test user, which requires an appropriate `storage` for our test user and probably a full `user` registered in the respective services. Has anybody experiences on doing this? @nextcloud/notes Please help 😃

jancborchardt:
Maybe @MorrisJobke or @LukasReschke have pointers here?

korelstar:
Yes, that would be nice, if somebody helps us, here.

.travis.yml
```diff
@@ -16,54 +15,19 @@ branches:
...
 
 before_install:
-  - export DISPLAY=:99.0
-  - export CHROME_BIN=chromium-browser
-  - sh -e /etc/init.d/xvfb start
-  - sudo apt-get update
-  - sudo apt-get install firefox
-  - wget https://raw.githubusercontent.com/owncloud/administration/master/travis-ci/before_install.sh
+  - wget https://raw.githubusercontent.com/nextcloud/travis_ci/master/before_install.sh
   - bash ./before_install.sh $APP_NAME $CORE_BRANCH $DB
```
nickvergessen: Use ` - . ./before_install.sh $APP_NAME $CORE_BRANCH $DB` (replacing `bash` with `.`) to fix unit tests on php 7.0 and 7.1

nickvergessen:
I pushed some commits, lets see what travis says

korelstar:
Thanks for helping out, @nickvergessen !
Unfortunately, travis is still red ... :-(

nickvergessen:
Sorry no idea anymore, it seems the mountpoint of the users home is not added to the database. This means getById() can not query for the file.

I would assume that @icewind1991 could help with that.

korelstar:
Thank you, anyway!

jancborchardt:
Same here, any updates @korelstar @nickvergessen? (ref https://github.com/nextcloud/notes/pull/36)

Also cc @Henni @camilasan

korelstar:
No news, here. Some help would be great :-)

jancborchardt:
cc @icewind1991 @rullzer @MorrisJobke  – any help appreciated :)

korelstar:
Closing, since we have no progress for a very long time and the code is outdated in the meanwhile.

PR #60, opened 2017-02-05
Tiny changes in README: move mobile clients into text

korelstar:
See the discussion of 1f9bd60de38e57417a51e7fcef2a0a8d819fbbc5.
Now, I've put all mobile clients into the summary. Therefore, they are better visible (GitHub and appstore) and no client is privileged.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -5,7 +5,9 @@
     <name>Notes</name>
     <name lang="de">Notizen</name>
     <summary>Distraction-free notes and writing</summary>
-    <description><![CDATA[The Notes app is a distraction free notes taking app. It supports formatting using [Markdown](https://en.wikipedia.org/wiki/Markdown) syntax. Notes are saved as files in your Nextcloud, so you can view and edit them with every Nextcloud client. Furthermore, a separate [RESTful API](https://github.com/nextcloud/notes/wiki/API-0.2) allows for an easy integration into third-party apps such as the [Android client](https://github.com/stefan-niedermann/nextcloud-notes). Further features include marking notes as favorites and future versions will provide categories for better organization.]]></description>
+    <description><![CDATA[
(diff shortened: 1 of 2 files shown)


korelstar:
I've also changed the `max-version` for ownCloud, since the tests fail otherwise.

appinfo/info.xml
```diff
@@ -22,7 +24,7 @@
     <repository type="git">https://github.com/nextcloud/notes.git</repository>
     <screenshot small-thumbnail="https://raw.githubusercontent.com/nextcloud/screenshots/master/apps/Notes/notes-thumbnail.jpg">https://raw.githubusercontent.com/nextcloud/screenshots/master/apps/Notes/notes.png</screenshot>
     <dependencies>
-        <owncloud min-version="8.1" max-version="9.2" />
+        <owncloud min-version="8.1" max-version="10" />
```
MariusBluem: Do we plan to support and publish this app for Owncloud 10 🤔 @Henni

korelstar:
Why not, if it works? It can be removed when it will be broken sometime...

appinfo/info.xml
Henni: Probably not. But shouldn't break for now. So let's keep it in there.

appinfo/info.xml
```diff
@@ -5,7 +5,9 @@
...
+    <description><![CDATA[
+The Notes app is a distraction free notes taking app. It supports formatting using [Markdown](https://en.wikipedia.org/wiki/Markdown) syntax. Notes are saved as files in your Nextcloud, so you can view and edit them with every Nextcloud client. Furthermore, a separate [RESTful API](https://github.com/nextcloud/notes/wiki/API-0.2) allows for an easy integration into third-party apps (currently, there are notes apps for [Android](https://github.com/stefan-niedermann/nextcloud-notes) and [iOS](https://github.com/owncloud/notes-iOS-App) which allow convenient access to your Nextcloud notes). Further features include marking notes as favorites and future versions will provide categories for better organization.
+]]></description>
```
Henni: I believe we should be able to remove the CDATA tag here. But this can be done separately.

PR #82, opened 2017-04-13
Fix Travis

korelstar:
I think this should fix the build on Travis.


korelstar:
Hmpf. Didn't worked, the error is now at another point. Maybe someone else have an idea.

korelstar:
I'm closing this, since it didn't worked. I still hope that CI will work again at sometime ...

PR #87, opened 2017-05-05
add API to change the category

korelstar:
Since #6 introduced notes from subdirectories and exposes this in the 3rd-party API as `category` (see also #8), we've implemented read-only categories in the Android client (see stefan-niedermann/nextcloud-notes#181). However, there is the demand to change the category from the client.

Therefore, I extended the API to be capable of doing this (fully backward compatible). The implementation on the client-side is already finished (for now in a simple approach, this will be improved later) and the integration can be tested: stefan-niedermann/nextcloud-notes#210

Of course, the web app should be able to display and change categories, too. But my AngularJS skills are limited and there is no agreement on how to layout this new feature. However, this implementation is realized in the `notesservice`, so it can be used by the web app, too.

### DIFF
--- a/service/notesservice.php
+++ b/service/notesservice.php
@@ -15,6 +15,7 @@ use OCP\Files\FileInfo;
 use OCP\IL10N;
 use OCP\Files\IRootFolder;
 use OCP\Files\Folder;
+use OCP\ILogger;
 
 use OCA\Notes\Db\Note;
 
@@ -27,14 +28,20 @@ class NotesService {
 
     private $l10n;
     private $root;
+    private $logger;
+    private $appName;
 
     /**
      * @param IRootFolder $root
      * @param IL10N $l10n
+     * @param ILogger $logger
+     * @param String $appName
      */
-    public function __construct (IRootFolder $root, IL10N $l10n) {
+    public function __construct (IRootFolder $root, IL10N $l10n, ILogger $logger, $appName) {
         $this->root = $root;
         $this->l10n = $l10n;
+        $this->logger = $logger;
+        $this->appName = $appName;
     }
 
 
@@ -162,7 +169,8 @@ class NotesService {
             if($currentFilePath !== $newFilePath) {
                 $file->move($newFilePath);
             }
-        } catch(\Exception $e) {
+        } catch(\OCP\Files\NotPermittedException $e) {
+            $this->logger->error('Moving this note to the desired target is not allowed. Please check the note\'s target category.', array('app' => $this->appName));
         }
 
         $file->putContent($content);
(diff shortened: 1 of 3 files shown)


service/notesservice.php
```diff
@@ -262,6 +280,19 @@ private function generateFileName (Folder $folder, $title, $extension, $id) {
...
+        $cleanedFolders = [];
+        foreach($folders as $folder) {
+            $folder = trim($folder);
+            $folder = preg_replace('/^\.+\s*/', '', $folder);
+            if(!empty($folder)) {
+                $cleanedFolders[] = $folder;
+            }
+        }
+        return implode('/', $cleanedFolders);
+    }
```
Henni: We can drop this.
As discussed with @LukasReschke all paths containing `/../` are denied by core.

korelstar: I've just tested removing this method. Then, the core throws a `NotPermittedException`. This currently results in not accepting the whole change (i.e. also content changes) and in creating two (!) empty notes. In order to improve the user-experience (a respective category name could be entered by accident), we should handle this exception.
Of course, this can be easily done, but we have to think about, which behavior is desired:

a) Remove problematic parts from the new category name and perform changing the category to this adjusted name (my current approach).
b) Don't prepare category name; check for Exception by core; and ignore category change if Exception happens.

From the user perspective, I would prefer a). What do you think?

Henni: This should only occur with a comment containing `/../` right? So I really doubt that any user would enter such a category. I would probably go with b.

korelstar: Okay, I've removed the category name preparation and introduced an exception handling. Please review again, @Henni

service/notesservice.php
```diff
@@ -138,15 +138,31 @@ public function update ($id, $content, $userId, $mtime=0) {
...
+
+            // assemble new file path
+            $newFilePath = $basePath . '/' . $this->generateFileName($folder, $title, $fileExtension, $id);
+
+            // if the current path is not the new path, the file has to be renamed
+            if($currentFilePath !== $newFilePath) {
+                $file->move($newFilePath);
+            }
+        } catch(\Exception $e) {
         }
```
Henni: Can we specify a stricter Exception than the generic `\Exception`?
Also it would be great to at least give some kind of feedback in form of a server log entry.

korelstar: Okay, that's reasonable. I will change this.

korelstar: done; ready for merging, now :smiley:

Henni:
Awesome job @korelstar !

stefan-niedermann:
Great work as always! So i guess we can continue https://github.com/stefan-niedermann/nextcloud-notes/pull/210 as soon as it has been tested with combination old/new?

korelstar:
Thanks :-)
Please test the branch of the Android app. Maybe, there are some cosmetic issues.

PR #88, opened 2017-05-06
improve title generation

### LINKED ISSUE #55: Note renaming from within app
Yesterday I installed this app and created some test notes. What was very disappointing to me is that one cannot rename a note from inside the app, but has to do it in the Notes folder. This is not very intuitive.

korelstar:
I've improved the generation of the note's title respective filename. Now, much more markdown markups as well as white spaces are removed. The approach is based on a similar implementation in the Android client: stefan-niedermann/nextcloud-notes#143

This will fix #55.

### DIFF
--- a/service/notesservice.php
+++ b/service/notesservice.php
@@ -121,22 +121,7 @@ class NotesService {
         $notesFolder = $this->getFolderForUser($userId);
         $file = $this->getFileById($notesFolder, $id);
         $folder = $file->getParent();
-
-        // generate content from the first line of the title
-        $splitContent = preg_split("/\R/", $content, 2);
-        $title = $splitContent[0];
-
-        if(!$title) {
-            $title = $this->l10n->t('New note');
-        }
-
-        // prevent directory traversal
-        $title = str_replace(array('/', '\\'), '',  $title);
-        // remove hash and space characters from the beginning of the filename
-        // in case of markdown
-        $title = ltrim($title, ' #');
-        // using a maximum of 100 chars should be enough
-        $title = mb_substr($title, 0, 100, "UTF-8");
+        $title = $this->getSafeTitleFromContent($content);
 
         // generate filename if there were collisions
         $currentFilePath = $file->getPath();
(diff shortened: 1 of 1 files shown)


service/notesservice.php
```diff
@@ -196,6 +181,33 @@ public function delete ($id, $userId) {
...
 
+    private function getSafeTitleFromContent($content) {
+        // prepare content: remove markdown characters and empty spaces
+        $content = preg_replace("/^\s*[*+-]\s+/m", "", $content); // list item
+        $content = preg_replace("/^#+\s+(.*?)\s*#*$/m", "$1", $content); // headline
+        $content = preg_replace("/^(?:=*|-*)$/m", "", $content); // separate line for headline
+        $content = preg_replace("/(\*+|_+)(.*?)\\1/m", "$2", $content); // emphasis
+        $content = preg_replace("/^\s+/m", "", $content); // whitespace at beginning
+        $content = preg_replace("/\s+$/m", "", $content); // whitespace at end
+        $content = trim($content);
```
Henni: Trimming should remove whitespace both from the beginning and the end of the string. So why use the two `preg_replace` above?

service/notesservice.php
```diff
@@ -196,6 +181,33 @@ public function delete ($id, $userId) {
...
+
+        // generate content from the first line of the title
+        $splitContent = preg_split("/\R/", $content, 2);
+        $title = $splitContent[0];
```
Henni: how about moving this part up and just doing all the `preg_replace` stuff on the first line?

service/notesservice.php
```diff
@@ -196,6 +181,33 @@ public function delete ($id, $userId) {
         $file->delete();
     }
 
+    private function getSafeTitleFromContent($content) {
+        // prepare content: remove markdown characters and empty spaces
+        $content = preg_replace("/^\s*[*+-]\s+/m", "", $content); // list item
+        $content = preg_replace("/^#+\s+(.*?)\s*#*$/m", "$1", $content); // headline
+        $content = preg_replace("/^(?:=*|-*)$/m", "", $content); // separate line for headline
+        $content = preg_replace("/(\*+|_+)(.*?)\\1/m", "$2", $content); // emphasis
```
Henni: RegEx magic ✨
I still have to thoroughly test this stuff.

service/notesservice.php
korelstar: The advantage of doing the replacements first is that removing markups can produce an empty line which would result in an empty title. For example, if the first line consists of some characters that form a markdown code that is removed by the regexps, only some whitespaces are left. I wanted to prevent this, since an empty title is useless.

service/notesservice.php
korelstar: Yes, I can do that. I took the separate trimming from the Android app, since we had to remove markups for the whole content (the list of notes shows a small snippet of the note, beside the title).

service/notesservice.php
Henni: @privatereese do you want to review this?

service/notesservice.php
```diff
@@ -196,6 +181,31 @@ public function delete ($id, $userId) {
         $file->delete();
     }
 
+    private function getSafeTitleFromContent($content) {
+        // prepare content: remove markdown characters and empty spaces
+        $content = preg_replace("/^\s*[*+-]\s+/m", "", $content); // list item
+        $content = preg_replace("/^#+\s+(.*?)\s*#*$/m", "$1", $content); // headline
```
Henni: I wasn't aware of closing heading markup. I believe this currently leads to issues in the web editor. But this is unrelated to this PR.

service/notesservice.php
```diff
@@ -196,6 +181,31 @@ public function delete ($id, $userId) {
...
+        $content = preg_replace("/^(?:=*|-*)$/m", "", $content); // separate line for headline
```
Henni: Why do you explicitly use a non-capturing group? Does this have any advantages as opposed to `/^(=*|-*)$/m`.

Also shouldn't we write this as: `/^(=+|-+)$/m`? Otherwise empty lines match. Or was this intended?

korelstar: I think, you're right. It's a long time ago, when I wrote those regular expressions. Therefore, I actually don't remember why I chose that variant. I will test your suggestion in detail and change it if it works.

korelstar: @Henni , you were right. I've change the expression. Ready for merge?

PR #95, opened 2017-05-26
API speed-up (ETag, Last-Modified)

korelstar:
This is a rework of #51. It's based on the comments made for the old PR.

The goal of this PR is to enhance the REST-API for 3rd-party (mobile) apps in order to reduce the data that has to be transferred during synchronization. Therefore, only changed notes should be transferred (see also #49 for the full motivation).

## API-Changes

**1. Don't transfer data if nothing has changed (`ETag`/`If-None-Match`)**

The server now sends the HTTP header field `ETag`, which is a hash value of the server response. After processing the server response, clients should save this value locally and send the HTTP header `If-None-Match` using the saved value in their next request to the server. If the server wants to response with the same data, it will send `304 Not Modified` without any data instead. This behavior is conform to the HTTP standard and fully backward compatible.

**2. Only transfer changed data, if something has changed (`pruneBefore`)**

If only a single note is changed (edited or deleted), we nevertheless don't want to transfer all notes again. Therefore, the server response includes the HTTP header `Last-Modified`. After processing the server response, clients should save this value locally and use the parameter `pruneBefore` in their next request to the server. Then, the server prunes all notes which where changed before this date (i.e. only their `id` is transferred). With this approach, clients are able to update changed notes and also remove deleted notes while reducing the transferred data size significantly. This behavior is fully backward compatible.

## Implementation background

The API changes are quite simple but very effective. However, the implementation is much more complex, since it's not that easy to get the *real* last modified date from a note (see https://github.com/nextcloud/notes/pull/51#issuecomment-279129420). Therefore, I introduced a database table (`notes_meta` with respective Mapper and Service), which tracks this information (see also https://github.com/nextcloud/notes/pull/51#issuecomment-279206258 for the rationale). It is only updated, when there is a request using the API. This means, that the detected change-date is not the real change-date, but it is the first time when the changed value is used -- therefore it is sufficient for the goal of this PR.

## Review

Please give feedback, @BernhardPosselt, @Henni, @nextcloud/notes! For the Android client, I've made a client-side implementation, which uses this enhanced API: stefan-niedermann/nextcloud-notes#220


db/metamapper.php
```diff
@@ -0,0 +1,29 @@
...
+
+	public function getAll($userId) {
+		$sql = 'SELECT * FROM `*PREFIX*notes_meta` WHERE user_id=?';
+		return $this->findEntities($sql, [$userId]);
+	}
+
+	public function get($userId, $fileId) {
+		$sql = 'SELECT * FROM `*PREFIX*notes_meta` WHERE user_id=? AND file_id=?';
+		return $this->findEntity($sql, [$userId, $fileId]);
+	}
```
korelstar: It would be nice, if there would be a way to do this. But I didn't find one (see https://github.com/nextcloud/notes/pull/51#issuecomment-279129420 for some use cases which have to be considered). If you have any idea on how to improve this, please give feedback. But also after the discussion with @BernhardPosselt, I don't see another way.

Henni: So `last-modified` inside the response header and `mtime` of the actual note (file) are independent values, right? The header value will be stored in the database and the mtime is directly set on the note.

But would it hurt to generate the ETag for every response instead of caching it in the database?

korelstar: > So `last-modified` inside the response header and `mtime` of the actual
> note (file) are independent values, right? The header value will be
>stored in the database and the mtime is directly set on the note.

Yes, mtime of the file is set by the client ("application-level" change time) and last-modified by the server (more technical level)

> But would it hurt to generate the ETag for every response instead of
> caching it in the database? 

This is already the case. The ETag is used to detect if the file has been changed. But the goal is to update "last-modified" of that file *if* it has changed. Only for this, we need the database.

Henni: But what is the ETag field in the database used for then?

korelstar: It is used to detect whether a note was changed (newly calculated ETag of a note differs from the ETag in the database) or not (newly calculated ETag is equal to the one in the database). In the first case, `last-modified` is updated to `now`. Based on this, the note will be fully transferred to all clients on their next request, even if they use `pruneBefore` to reduce the size of the response.

Please note, that this ETag is not equal to the file based ETag of nextcloud/server since it includes all attributes of a note, e.g. favorite tag (and the ETag in the response header of a list response is a completely different one).

korelstar: Any news on this? Can it be merged, now?

Henni: I'm fine with this.

Henni:
@korelstar awesome job as always!

PR #98, opened 2017-06-04
Nextcloud 13 compatibility

jancborchardt:
Compatibility with the new Nextcloud development version:

- adjust the max-version

- remove the lang="de" app name. App names are translated in Transifex, no? It also resulted »Array« to be displayed instead of the app name in the upgrade screen:
screenshot from 2017-06-04 22-48-28

Please review @nextcloud/notes

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -25,7 +25,7 @@ The Notes app is a distraction free notes taking app. It supports formatting usi
     <screenshot small-thumbnail="https://raw.githubusercontent.com/nextcloud/screenshots/master/apps/Notes/notes-thumbnail.jpg">https://raw.githubusercontent.com/nextcloud/screenshots/master/apps/Notes/notes.png</screenshot>
     <dependencies>
         <owncloud min-version="8.1" max-version="10" />
-        <nextcloud min-version="9" max-version="12" />
+        <nextcloud min-version="9" max-version="13" />
     </dependencies>
     <ocsid>174554</ocsid>
 </info>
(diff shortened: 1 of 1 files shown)


Henni:
👍 for nc 13

Not sure about the translation fix though... I took this from the [documentation](https://nextcloudappstore.readthedocs.io/en/latest/developer.html#info-xml).
@BernhardPosselt can you give us a quick hint? Does the appstore translate app names with transifex?

korelstar:
Here is another app using a translated title:
https://apps.nextcloud.com/apps/ocr / 
https://github.com/janis91/ocr/blob/master/appinfo/info.xml

If this is not allowed it should not be shown in the documentation. Otherwise, this looks like a bug in the server core.

BernhardPosselt:
App names and other translations are taken from the info.xml

The store has nothing to do with transifex

BernhardPosselt:
That doesn't mean though that we can't use the translations directory however someone with more knowledge would need to provide more details. E.g. we could do something like that:

```xml
<info>
    <name use-l10n="true">Notes</name>
</info>
```

That would look up translations in the l10n directory. However we need to find a mechanism that ensures that this works in a similar fashion in core.

jancborchardt:
As shown above it doesn't work in the server as of the current development version, so until it's fixed we should remove it here. Ok? :)

korelstar:
@jancborchardt
Did you reported this issue to the respective repository (nextcloud/server or nextcloud/updater, I think), so that this issue can be fixed?

korelstar:
@BernhardPosselt
Sounds like a good approach! This seems to be much better than the current approach :-)
Maybe, this should be discussed at a more central location, e.g. in the repository of nextcloud/server or nextcloud/appstore.

jancborchardt:
Feel free to open an issue, but since this problem exists with a released Nextcloud version, we should fix this here too. ;)

korelstar:
I don't have the knowledge about this bug, since it didn't appear on my installation. So please (!) report that bug (that you (!) found) by yourself and do not try to doctor on symptoms, only!

PR #101, opened 2017-06-21
More natural heading-1 size

andreasjacobsen93:
Decreased size of heading-1 to make it match the other headers more.
I felt it was way too big.

### DIFF
--- a/css/notes.css
+++ b/css/notes.css
@@ -153,7 +153,7 @@
 	font-weight: 600;
 }
 .CodeMirror .CodeMirror-code .cm-header-1 {
-    font-size: 40px;
+    font-size: 28px;
     margin: 50px 0 20px;
     line-height: 120%;
 }


korelstar:
IMHO something in the middle (32px?) would be fine.
/cc @jancborchardt

jancborchardt:
Can you show a before-after comparison screenshot? :)

The h2 and h3 should match the 20px and 15px defined in https://github.com/nextcloud/server/blob/master/core/css/apps.scss, and the h1 here should look good with that. :)

andreasjacobsen93:
headers

jancborchardt:
Looks great! :+1: What do you think @nextcloud/notes?

jancborchardt:
By the way @andreasjacobsen93 @korelstar it would be great to have you at the Nextcloud conference end of August in Berlin! https://nextcloud.com/conf/
Would be especially cool if you can give a lightning talk about your work on Nextcloud! :)

Henni:
Looks better!

Issue #44, opened 2017-01-05
Discussion: Style 'add note' as button


Henni:
Should we style the 'add note' action like a button? This would align notes to other apps like contacts:
![](https://cloud.githubusercontent.com/assets/6277619/19726700/745d175c-9b8c-11e6-8f78-d12336c9b7ee.jpg)

see also:
https://github.com/nextcloud/notes/pull/42#discussion_r94250427

cc @nextcloud/designers

skjnldsv:
Yes, we should aim to uniform designs :)

MariusBluem:
👍

jancborchardt:
I agree with this, yes. It makes the primary action clearer and we do this in Contacts & Mail as well. (And to some extent in Files and Calendar)

Henni:
Looks like we currently don't have a `.primary-button` style in core.
I think we should change that so that we can reuse it in contacts, mail and here in notes.

skjnldsv:
I completely agree.
Should be in apps.scss. A `#app-navigation > .primary-button-container > button.primary-button`?

- Mail has its button inside a div
- Contacts has it inside a custom element (which a class can be easily applied)
- Calendar has 3 buttons (previous, month and next) maybe we should uniform the `.primary-button-container` with some flexbox to be compatible with that too?

Henni:
I would ignore calendar and just define a simple primary button.

A wrapper element like `.primary-button-container ` shouldn't be necessary. And maybe we should even remove #app-navigation to allow primary buttons in the app-content.

Also: Do we only want button or also input[type=button]?

skjnldsv:
Since every app is using a wrapper, it could be useful to think twice about it.
I agree with you, this isn't necessary, but I like the idea of a unified buttonS wrapper wich include calendar.

jancborchardt:
We do have `.primary` though, as used by the log in button and the main button of the file picker, and in the Calendar app for event saving, and in Mail for send and reply button. :) So that should be used.

jancborchardt:
This is fixed, we now have a nice button there. :)

jancborchardt:
Thanks to @andreasjacobsen93 who fixed it in https://github.com/nextcloud/notes/pull/108 ;)

PR #103, opened 2017-06-27
Styled 'create note' as button

andreasjacobsen93:
As proposed here https://github.com/nextcloud/notes/issues/44, I changed the create new note button.

capture

NB. I changed the text to 'Create new note', so should probably be updated in transifex?

### DIFF
--- a/css/notes.css
+++ b/css/notes.css
@@ -8,13 +8,24 @@
     padding-bottom: 0; /* no settings bar, so no padding needed */
 }
 
+#app-navigation #note-add {
+    height: 30px;
+    margin-top: 11px;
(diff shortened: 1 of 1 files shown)


Henni:
Looks good so far.
I would change two things: Make the button full-width and left-align the text.
What do you think, @andreasjacobsen93 ?

cc @jancborchardt

jancborchardt:
Yep, absolutely agree with @Henni's proposed changes. :)

andreasjacobsen93:
Had to move to new PR https://github.com/nextcloud/notes/pull/108
Because of a problem with my fork

Issue #93, opened 2017-05-23
Special characters not work


cayetano:
Latin characters like á é í ó ú ñ etc... not work

Regards
Cayetano

korelstar:
Which version do you use? There is a bug fix in the area of charter encoding which isn't released yet. But you can test it if you use the master branch of this Git repository.

If the problem still occurs, please give more details what exactly doesn't work.

notklaatu:
Titling a note this:
``Pączki``
causes the notes app to hang. The note text file in $USER/Notes is labelled ``P``.

notklaatu:
nextCloud-clients can no longer sync, citing missing file ``P``. Disabling Notes, moving Note files to a local machine, deleting Notes from server, re-enabling notes, and then copying old notes back to server appears to fix the problem.

korelstar:
Ouh, I can confirm this bug. This is really uncomfortable, since it's really hard to delete the file with the bad character.

This definitely has to be fixed.

korelstar:
This should be fixed, now (in `master`). Please see #125.

PR #125, opened 2017-08-18
fix preg multibyte support

Loki3000:
$content and $title is multibyte strings. So preg_* functions must contain /u modificator.


korelstar:
Interesting, maybe this fixes also #93 ?

Currently, I'm not able to test this, but I will do in about a week...

Loki3000:
It's possible. I've get errors with some non-english strings.

korelstar:
Thank you very much for your contribution! This was an important bug fix :-)

PR #137, opened 2017-10-05
Fix error handling for save / manual save

### LINKED ISSUE #113: Allow to manually save notes
Currently, and in reference to all the other issues related to saving notes and losing notes, I propose the idea to implement a manual save method, either via a button or Ctrl+S, including a message showing "Note saved" (or, if it failed, an appropriate error message).

Any ideas on this? Could that be implemented?

korelstar:
I've fixed resp. implemented some (in my view) basic features regarding error handling / prevention of data loss. Some of them existed before, but stopped working (maybe at the switch to the new editor?), some of them are new.

- [x] Show if a note is not saved yet (by displaying a `*` next to the note title / word counter).
- [x] Show a warning (non-blocking red box) if saving the note has failed (disappears automatically, if the next save succeeds); fix #113.
- [x] Introduce manual save of the current note by pressing CTRL+S (then a loading icon appears during the save process).
- [x] Warn the user if there is a note which hasn't been saved yet when leaving the page.

/cc @nextcloud/designers


jancborchardt:
Nice, especially on the Ctrl-S hijack! :) I didn’t test but it sounds fine by me

templates/main.php
```diff
@@ -51,6 +51,7 @@
                 ng-class="{ active: note.id == route.noteId }">
                 <a href="#/notes/{{ note.id }}">
                     {{ note.title | noteTitle }}
+                    <span ng-if="note.unsaved">*</span>
```
Henni: Could be solved with `ng-class` on the link instead. This way the * or any other styling could be injected by css (::after).

PR #148, opened 2017-11-04
Sanitise the note's path

korelstar:
I've implemented some enhancements in order to let the notes app to be more fail-safe.

- [x] Remove characters in note's file path (i.e. note's title and category) that are forbidden on other operating systems (i.e. Windows). The problem is, that we can title a note `Math <> Philosophy` on a (linux) server, but when the user wants to synchronize the account with a Windows system, synchronization fails. Therefore, we should remove all characters that lead to failed synchronization (fixes stefan-niedermann/nextcloud-notes#228).
- [x] Renaming a note (e.g. title or category has changed) can lead to an exception. Until now, we only catched the `NotPermittedException` (user has not the right to write to a (shared) folder), but other exceptions can happen, too (e.g. `InvalidPathException` if the filename is invalid due to several reasons). Those are now catched, too. This allows the note content to be saved while although renaming has failed.
- [x] An example for when the `InvalidPathException` is thrown, is when using an emoji in the title and having a MySQL database without 4-byte support. The new function `sanitisePath` removes such illegal characters from the path in such a case (fixes stefan-niedermann/nextcloud-notes#237).

### DIFF
--- a/service/notesservice.php
+++ b/service/notesservice.php
@@ -214,8 +214,13 @@ class NotesService {
 
     // removes characters that are illegal in a file or folder name on some operating systems
     private function sanitisePath($str) {
-        $str = str_replace(['*', '|', '/', '\\', ':', '"', '<', '>', '?'], '', $str); // problematic characters
+        // remove characters which are illegal on Windows (includes illegal characters on Unix/Linux)
+        // prevents also directory traversal by eliminiating slashes
+        // see also \OC\Files\Storage\Common::verifyPosixPath(...)
+        $str = str_replace(['*', '|', '/', '\\', ':', '"', '<', '>', '?'], '', $str);
+
         // if mysql doesn't support 4byte UTF-8, then remove those characters
+        // see \OC\Files\Storage\Common::verifyPath(...)
         if (!\OC::$server->getDatabaseConnection()->supports4ByteText()) {
             $str = preg_replace('%(?:
                 \xF0[\x90-\xBF][\x80-\xBF]{2}      # planes 1-3
@@ -223,7 +228,9 @@ class NotesService {
               | \xF4[\x80-\x8F][\x80-\xBF]{2}      # plane 16
               )%xs', '', $str);
         }
-        $str = preg_replace("/^[\. ]+/mu", "", $str); // hidden files
+
+        // prevent file to be hidden
+        $str = preg_replace("/^[\. ]+/mu", "", $str);
         return trim($str);
     }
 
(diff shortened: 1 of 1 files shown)


service/notesservice.php
```diff
@@ -205,15 +212,32 @@ public function delete ($id, $userId) {
...
+    // removes characters that are illegal in a file or folder name on some operating systems
+    private function sanitisePath($str) {
+        $str = str_replace(['*', '|', '/', '\\', ':', '"', '<', '>', '?'], '', $str); // problematic characters
+        // if mysql doesn't support 4byte UTF-8, then remove those characters
+        if (!\OC::$server->getDatabaseConnection()->supports4ByteText()) {
+            $str = preg_replace('%(?:
+                \xF0[\x90-\xBF][\x80-\xBF]{2}      # planes 1-3
+              | [\xF1-\xF3][\x80-\xBF]{3}          # planes 4-15
+              | \xF4[\x80-\x8F][\x80-\xBF]{2}      # plane 16
+              )%xs', '', $str);
```
Henni: Would you mind adding some explanatory comments to this regex magic?

service/notesservice.php
```diff
@@ -222,9 +246,6 @@ private function getSafeTitleFromContent($content) {
             $title = $this->l10n->t('New note');
         }
 
-        // prevent directory traversal
-        $title = str_replace(array('/', '\\'), '',  $title);
```
Henni: Did you look at the core code and how it is prevented there?
I would prefer if we wouldn't have to implement this ourselves...

korelstar:
Thanks for reviewing this and the other PRs.

Some background information on this one:

I checked the core code for possible exceptions during rename. There, several types of checks are executed (see lib/private/files/Storage/Common.php, verifyPath()). However, if one of the checks fails, an Exception is thrown. Catching those exceptions is not sufficient for us, since renaming is then aborted and the note's title is left in its old state which confuses the user who didn't do anything wrong (e.g. used a unicode smiley in the note's first line).

Therefore, I took those checks and eliminate everything during title generation that leads to an Exception. The magic regexp comes from a check from core, too (eliminate 4byte Unicode characters if MySQL can't save them). With this approach, the user can write everything in the first line of a note and the app only uses suitable parts for the file name.

I think we have to go this approach in order to provide a good user experience. But, of course, I can improve the documentation of the respective code.

Henni:
@korelstar thanks for the clarification.
I would be happy with a short reference to the core code inside a comment for both code pieces.

korelstar:
done

Issue #24, opened 2016-12-02
Add setting to create new files with markdown file extension (.md)


eppfel:
I don't know, if this was discussed before, but new files are created with the .txt extension. If you want to edit the files directly in the `Notes` folder with another editor (e.g. [icewind1991/files_markdown](https://github.com/icewind1991/files_markdown) or on Desktop), the file format is not recognized.

Henni:
Should we introduce a setting for this? I think defaulting it to `.md` might irritate some users which are unable to open it on their system because they have never seen this extension. But I could be overthinking this.
@jancborchardt your opinion on this?

korelstar:
I vote definitely for a setting for this. `.md` is too nerdy to be the default.  :sunglasses:

phsc:
Another idea:
Instead of adding a setting, I would suggest to do it similar to the Files app, which adds .txt by default to the end of a new file.
1. Add .md by default to the end of a new note. If someone overwrites .md with .txt for whatever reason, accept it. Save the note with the according file extension. If someone adds another file extension, display an error like "File extension unsupported. Please use '.md' or '.txt'!".
2. Add an info box, explaining .md (for all the non-nerds): "The Notes app uses Markdown to save notes. The default file extension for Markdown is '.md'. If you prefer, you can also use '.txt' alternatively."

jancborchardt:
Sooo first off, @icewind1991’s Markdown editor should simply also recognize txt files. ;)

About the rest I need to think more, but .md should not be a default for sure. It’s only used and known by a tiny percentage of people.

eppfel:
So, I really would like to have the setting. Otherwise I will have to write a service that checks the Notes folder and replaces the extension.

And I am curios: Would users that don't know about `.md` really manually edit their notes with a text editor or would they have another notes app for mobile/desktop?

jancborchardt:
The problem with a setting is (as I also told @Henni) that it can’t be a global setting really. Because sometimes you might write markdown, sometimes just text.

Isn’t there a way we can detect as soon as markdown is used and rename? I mean, the title can also change at any point. And Markdown can be detected by `##` (single hash could be used for other things) or `[link](url)` or `**bolding**` or more.

korelstar:
Auto detection will not work always correctly, so there would be still the need for a manual setting.

I still think that a global setting would be fine (it would be definitely better than auto detection without manual correction per note), since people who are familiar with markdown and activate `.md` globally are still able to fill the files with text only.

[...]

PR #153, opened 2017-12-02
Change default file extention

lefuturiste:
Changed txt to md because it's markdown note, it's useful because we can, after download the note, or on windows client or whatever edit the notes with appropriated software

tested on nextcloud 12

### DIFF
--- a/service/notesservice.php
+++ b/service/notesservice.php
@@ -107,7 +107,7 @@ class NotesService {
         // check new note exists already and we need to number it
         // pass -1 because no file has id -1 and that will ensure
         // to only return filenames that dont yet exist
-        $path = $this->generateFileName($folder, $title, "txt", -1);
+        $path = $this->generateFileName($folder, $title, "md", -1);
         $file = $folder->newFile($path);
 
         return Note::fromFile($file, $folder);


marcelstoer:
The second thing that irritated me about this app was exactly this...looking forward to seeing this merged and released.

Orhideous:
@Henni, @LukasReschke — can you please merge this PR? It would be nice to open notes directly from another app (MarkDown App) later :)

sprnza:
Guys, please merge. I'd like to get rid of the annoying "code integrity check" error.

brantje:
Tested with NC13 and 14 no issues found.

lachmanfrantisek:
Hi all,

the setting would be nice, but what about merging this and leave the settings-part to another PR? 

It looks like the PR author is probably not interested in doing this and the change of the default extension is therefore blocked.

Thanks!

jancborchardt:
The issue for this is called "Add setting" https://github.com/nextcloud/notes/issues/24, and that's what we also really need in this case. Normally we do step-by-step developments, but here **changing the default would break it for any Windows user who does not know what md files are.** This is not something we can do, so the setting has to be there. Otherwise we would just swap from one issue to the next.

Any code contribution or alternatively [donation towards the issue on Bountysource](https://www.bountysource.com/issues/39724834-add-setting-to-create-new-files-with-markdown-file-extension-md?utm_campaign=plugin&utm_content=tracker%2F46751022&utm_medium=issues&utm_source=github) is much appreciated! :tada:

korelstar:
Closed due to inactivity and with respect to #223, which will be finished in the next days.

PR #167, opened 2018-02-12
Remove unneeded files

MorrisJobke:


jancborchardt:
@MorrisJobke is it because they are incomplete, or because the translations are elsewhere anyway?

jancborchardt:
So @Henni @korelstar @camilasan @stefan-niedermann can you review? :)

PR #184, opened 2018-04-11
Moves error messages and word count to top.

camilasan:
Because I have lost notes before because I didn't notice the errors on the bottom of the page.
I also added some extra padding and margin in the error message.

notes-with-errors
notes-with-no-errors

cc @jancborchardt @stefan-niedermann


jancborchardt:
Just for the record, @camilasan @stefan-niedermann and me talked, and we will:
- Put the word count back to the bottom
- For the error notification we should use the standardized `OC.Notification.showTemporary(message);` (like the Mail app does too for example, and it doesn’t have big red scary elements ;)
- We should be a bit more informative: Instead of »Saving failed«, maybe »Server not reachable, note could not be saved« or »You are not connected to a network, saving will not work«? (in that case just .show and not .showTemporary)

camilasan:
I will do changes as requested as soon as I can. I have already spent way too much time replying to the Github issues in the client. Sorry.

jancborchardt:
So @camilasan and I talked about this again and came to the conclusion that: Cause we don’t want the wordcount at the top and we already have a standard way to show errors, it’s probably easiest to scrap this pull request and instead do the error display using the `showNotification` function. :)

PR #187, opened 2018-04-13
explicit width of editable cell

madbob:
On mobile, the contenteditable div ends up having a fixed (and small) width [due global CSS](https://github.com/nextcloud/server/blob/123d9f0ce960a3ed0010bc8bcd86273d25ce2d50/core/css/inputs.scss#L30).


korelstar:
Thanks!

Issue #186, opened 2018-04-13
Improve notification of "decryption failed exception"


KimTheFirst:
Firstly, this error was triggered by myself corrupting one of the files whilst doing something, so the corruption itself is not the issue. 

The issue here is that if this was indeed caused by the corruption of a single file within the notes app, the app now does not function at all. Maybe the file should show with ("decryption failed") or something?

KimTheFirst:
This is because the exception "DecryptionFailedException " is being thrown by the encryption library, and notes is not catching that exception.

It would be possible to catch the exception DecryptionFailedException in notesservice.php and then do something with it. I can see how this might encourage users to ignore what might potentially be a very serious problem, but personally I think perhaps having the app working with errors might be better than having it fail completely.

Line268:
```
    private function getFileById ($folder, $id) {
        $file = $folder->getById($id);
```

KimTheFirst:
I think there are some other potential exceptions too & definitely a few other places they would have to be caught / used.

korelstar:
Thanks for your analysis, @KimTheFirst !

Providing an error message to the frontend would be the best (see https://github.com/nextcloud/notes/pull/184#issuecomment-380893531 for some ideas on how to present error messages), but I think it's fine to start with catching the Exceptions and writing something meaningful to the log. Could you help us with implementing this, @KimTheFirst ? A pull-request is very appreciated :-)

jancborchardt:
Copying over my comment from https://github.com/nextcloud/notes/pull/188#issuecomment-387909164 :)

> can you adjust the style so it is not a modal, but an emptycontent view? Like in Files when you have an empty folder, with a nice icon (icon-edit), a h2 (»Error loading note«) and a short error description.

PR #188, opened 2018-04-14
Prevent app from crashing upon encountering faulty or missing file.

KimTheFirst:
#186 
-Informative errors compliant with NC standard.
-Dud files show in sidebar with error.

-Unit tests not extended to cover new function
-db/note.php error implementation could probably be better / less hacky


controller/pagecontroller.php
```diff
@@ -62,15 +63,21 @@ public function index() {
...
         try {
             $this->notesService->get($lastViewedNote, $this->userId);
+            $errorMessage=null;
         } catch(NoteDoesNotExistException $ex) {
             $lastViewedNote = 0;
+            $errorMessage="The last viewed note cannot be accessed. ".$ex->getMessage();
+        } catch(DecryptionFailedException $ex){
+            $lastViewedNote = 0;
+            $errorMessage=$ex->getMessage();
         }
```
korelstar: Can we merge all exceptions to a single `catch(\Exception)` here? Then we are prepared if a specific exception type is missing.

css/notes.css
```diff
@@ -19,7 +19,9 @@
 #app-navigation .active a {
 	padding-right: 70px;
 }
-
+#app-navigation > ul > li.note.has-error a{
+    color:#f00;
+}
```
korelstar: This doesn't work for me, since the `<li>` has not the class `note`. The CSS selector should be `#app-navigation > ul > li.has-error a` instead.

db/note.php
```diff
@@ -67,6 +94,28 @@ public static function fromFile(File $file, Folder $notesFolder, $tags=[]){
...
+     * @return static
+     */
+    public static function fromException($message,File $file,Folder $notesFolder,$tags=[]){
+        $note = new static();
+        $note->setId($file->getId());
+        $note->setErrorMessage($message);
+        $note->setError(true);
+        $note->setContent($message);
+        $note->setModified(null);
+        $note->setTitle($message); // remove extension
```
korelstar: The title should be the same like in the non-error case (note's title from filename). Or is the filename encrypted, too?

db/note.php
```diff
@@ -67,6 +94,28 @@ public static function fromFile(File $file, Folder $notesFolder, $tags=[]){
...
+     * @param File $file
+     * @return static
+     */
+    public static function fromException($message,File $file,Folder $notesFolder,$tags=[]){
+        $note = new static();
+        $note->setId($file->getId());
+        $note->setErrorMessage($message);
+        $note->setError(true);
+        $note->setContent($message);
+        $note->setModified(null);
```
korelstar: The Android app (accidentally) ignores notes with `modified=null`. Hence, note errors will not show up in the Android app. This may be intended (otherwise, the Android app has to deal with such special notes). However, the erroneous notes can only be detected in the web app (which is not a degradation to the current situation!). Is this what we want? (I think it's okay).
/cc @stefan-niedermann

js/app/services/notesmodel.js
```diff
@@ -27,6 +27,10 @@ app.factory('NotesModel', function () {
             return this.notes;
         },
         get: function (id) {
+            if(this.notesIds[id].error) {
+                OCdialogs.alert(this.notesIds[id].errorMessage,'Error!');
```
korelstar: This leads to a warning in gulp: `'OCdialogs' is not defined.`

Why do you aren't using the official Nextcloud way (using `OC.Notification` like in `appcontroller.js`) to present the error?

db/note.php
```diff
@@ -31,21 +32,43 @@
...
+
+        $id=$file->getId();
+
+        try{
+            $note=self::fromFile($file, $notesFolder, $tags=[]);
+        }catch(FileNotFoundException $e){
+            $note = self::fromException('File error: ('.$file->getName().') '.$e->getMessage(), $file, $notesFolder, array_key_exists($id, $tags) ? $tags[$id] : []);
+        }catch(DecryptionFailedException $e){
+            $note = self::fromException('Encryption Error: ('.$file->getName().')'.$e->getMessage(), $file, $notesFolder, array_key_exists($id, $tags) ? $tags[$id] : []);
+        }
```
korelstar: Can we add a `catch(\Exception)` here or merge all exceptions to a single catch here? Then we are prepared if a specific exception type is missing.

jancborchardt:
@KimTheFirst nice! :) Could you post some screenshots also for quicker design review?

db/note.php
KimTheFirst: It currently shows the exception / error along with the filename in both the title and content fields ie "Decryption Error: (somefile.txt) exception blahblah". I guess this function is also fulfilled by the error popups so it seems a bit redundant, the user can just click to see the error.

controller/pagecontroller.php
KimTheFirst: Sure. I think the rationale for doing it this way was to append messages to the first exception which often returns with a blank message, but I can just prefix to every message.

js/app/services/notesmodel.js
KimTheFirst: OCDialogs is in nextcloud core, but I feel like perhaps OC.dialogs is what is intended for apps to use. I think they both map to the same object. I'll update the reference.

OC.dialogs produces file error messages consistent with the file error messages in nextcloud core, so in my opinion this method is preferable

KimTheFirst:
error

db/note.php
KimTheFirst: I'm not near an Android device, but I'll check this out when I am. The error messages have more or less the same structure as the ordinary notes, so I think it might be possible for the Android app to recognise them.

korelstar: Yes, if `modified!=null`, then the error message looks like a real note. Hence, the app allows the user to open and edit the error message. This is not useful, since saving will obviously fail. I see two posibilities:

1. Stay with `modified=null`, no changes in the Android app. Faulty notes are not shown to the user.
2. Set `modified=filemtime()` and change the Android app, so that notes with error are shown read-only.

/cc @jancborchardt

camilasan:
Is it ok to merge it? @jancborchardt

camilasan:
I will take the blame :)

jancborchardt:
Sorry for being late, here’s my comment :)

@KimTheFirst when does this show up? On Notes app start, or on opening the note? Cause it should only show up on opening the note. 

And can you adjust the style so it is not a modal, but an emptycontent view? Like in Files when you have an empty folder, with a nice icon (icon-edit), a h2 (»Error loading note«) and a short error description.

KimTheFirst:
Hi @jancborchardt 

It shows up when the note is opened, meaning that it also shows when the app loads if the last note opened is now corrupt or inaccessible. My original motivation was that if the last opened file was corrupt then this effectively blocked the app from being used as it would just keep trying to re open that file.


I'm a bit busy at the moment; I will send PR with the requested design change in due course.

jancborchardt:
@KimTheFirst cool, thanks a lot! :)

PR #207, opened 2018-06-15
Make notes path configurable

danielmorlock:
I introduced a user setting where a user can set a custom path for its notes. See "Settings" -> "Additional Settings". 

Feedback is welcome!


controller/settingscontroller.php
```diff
@@ -0,0 +1,48 @@
...
+		$this->root = $rootFolder;
+	}
+
+    /**
+     * @NoAdminRequired
+     */
+    public function setNotesPath($notesPath) {
+	    $uid = $this->userSession->getUser()->getUID();
+
+	    $path = '/' . $uid . '/files' . $notesPath;
```
korelstar: I think we should sanitize `$notesPath` before using it here. However, I wasn't able to attack this. Does the Files API checks for directory traversals?

danielmorlock: I can't find anything about this in the dev docs. The closest info I've found is this: https://github.com/nextcloud/server/blob/stable13/lib/public/Files/Folder.php#L104
Since the path was not "manually" sanitized before my changes, I did not sanitize them, too and I assume that this is done internally. But you're right, this should be verified - but I think I'm not the right person to do so :)

danielmorlock:
@jancborchardt I've introduced the app settings in the bottom left. Further I did some work on making the settings management more abstract in order to support possibly other app settings next to "notesPath".

korelstar:
Thanks! That looks good to me:

grafik

grafik

----------------------

The personal settings page is obsolete now and should be removed. I will do a code review again, then (if @jancborchardt has no change requests).

danielmorlock:
@korelstar I've removed the obsolete settings page.

korelstar:
Thanks, @danielmorlock !

One last thing:

After changing the setting, nothing (visible) happens. It would be helpful, if the notes app would reload in order to make the new path effective. Can you add this functionality, please? After that I will merge this PR (if @jancborchardt doesn't intervene until then).

jancborchardt:
Is it also possible to make the setting apply without a reload? :)

korelstar:
Indeed, that would be nicer. That would require some more work. I think you should open a new issue for that.

HarleyStone:
In looking at the Android app, I'm unable to locate the Advanced options within settings. Currently running 0.27.1 Thanks.

stefan-niedermann:
Hi @HarleyStone,

this option is only available via the web interface.

@korelstar is there a REST-API for this setting?

HarleyStone:
> Hi @HarleyStone,
> 
> this option is only available via the web interface.
> 
> @korelstar is there a REST-API for this setting?

Ahhh. I see now. Thanks.

korelstar:
> @korelstar is there a REST-API for this setting?

No. Do you think we need this? I would say, that a setting in the server app is sufficient, isn't it?

stefan-niedermann:
Not sure. Noone asked until now and it is something you will usually do max. 1 time, so.... No.

If there was an API, i might have implemented a setting, but it's not worth the effort i think.

Issue #130, opened 2017-08-28
Offset Error for `#` Chars in Headings


ocelotsloth:
Markdown provides that you can either use one set of hashes leading a header line, but it also says you can have the same number of hashes trailing.

I don't personally see why one would use this particular feature of the markup language, but to stay in spec it might be worth finding a way to disable the offset for the second set of hashes.

![2017-08-28](https://user-images.githubusercontent.com/9255772/29794112-a2f8f24c-8c14-11e7-8b2f-93c1079c9621.png)

korelstar:
Fixed by @DMuckerman in #208

PR #208, opened 2018-06-19
Remove margin-left offset for any headers after the first in a line

DMuckerman:
This should address #130, by changing the width and margin-left on any headers after the first in a line.


korelstar:
Thank you very much for your contribution, @DMuckerman ! :+1:

Issue #43, opened 2017-01-04
Search is not where expected


stefan-niedermann:
Hi @Henni ,
i recently updated the app and was happy to see the search functionality.
Though i didn't expect it in the left sidebar but in the top right corner next to the user name where it is in the files, contacts and calendar app, too.

So this is a request to move it to the well-known place :)

PS.: Works great! :)

cc @nextcloud/designers

DirkR:
Hi @stefan-niedermann,
look at the "search function" as a filter: it filters the list of articles matching the search pattern in note title as well as body. Thus that's the correct place.

We might discuss, if the glass icon is the best icon for this job (=field). I would prefer a funnel. And one could argue, that the input field is somewhat hidden thanks to the "flat" design. A really visible input field might be better/more helpful. But the place is correct (IMHO).

Henni:
@jancborchardt would love to have your feedback on this, as we talked about this a while ago.

Bugsbane:
> it filters the list of articles matching the search pattern in note title as well as body.

I would expect that most users would think of this as "Search" and thus look for it with the traditional search icon and UI location, as the original poster did on this issue.

eppfel:
..and even if we provide full-text search later, the main use-case is finding the right note.

jancborchardt:
So, thinking about this it seems that uniformity with the other apps is more important than the value of experimenting with a new position for the search.

That’s a separate topic though still. With other considerations like accessibility on mobile – if it’s in the top of the sidebar, searching will be difficult to access and require a two-step process, if it’s discovered at all. Unlike the top right spot which people are used to from Android and also iOS to a certain extent.

So @Henni if you agree then the sensible solution might be to use the normal search field position after all.

splitt3r:
Related to this it would be nice if the "Add note" button would look the same as the "Add contact" button.

![add_note](https://cloud.githubusercontent.com/assets/1869584/22699145/ea818cbc-ed56-11e6-9843-87d2a4c4405e.png)

stefan-niedermann:
@jancborchardt @camilasan and i discussed this topic again in a design review at the hackweek and we came to the conclusion that even if it is technically *filtering* and not *searching* (what i do not aggree), the input should be on the top right position.

skjnldsv:
@stefan-niedermann I'll introduce a new standard for the search button soon :)

jancborchardt:
@skjnldsv anything we could implement here already? :)

[...]

PR #210, opened 2018-06-23
Categories in navigation

korelstar:
Next step to provide categories to the frontend (see #8).

This PR is only about the navigation on the left side. The goal is to provide a list of existent categories with the possibility to filter the list of notes by the currently selected category. The design should orient on similar selectors of other Nextcloud apps as well as on the Android app. Changing the current note's category must be done in a separate PR (see #228).

## Tasks done
- [x] show list of categories in the navigation bar (left side)
  - generate list from current notes
  - show only first level of categories
- [x] make these entries selectable
- [x] filter list of notes by current selection
- [x] show the number of notes per category as counter
- [x] when using the "New note" button, the new note is assigned automatically to the currently selected category
- [x] don't show favorites as navigation item (see discussion in the comments)
- [x] don't auto-close navigation after clicking on category (only on small screens)
- [x] group notes by category, if there is a sub-category (like in the Android app)

## Feedback
Please feel free to test and give feedback, especially from the design perspective. @nextcloud/designers @jancborchardt 
I will also provide some screenshots ...


korelstar:
# Screenshots of the current implementation

## Default view:
grafik


## ... after clicking on "Categories" (collapsible):
grafik


## ... after clicking on the category "Test":
grafik

korelstar:
Oh, and please note: currently, only the first level of categories is shown in the navigation. E.g., if there are notes in the folders `Test/A/`, `Test/B/` and `Moin`, then the list of categories in the navigation is `Test`, `Moin`.

jancborchardt:
Hmm, the 2-level navigation is a bit confusing, and not fully necessary:
- recent is the default sorting
-favorites should be sorted up top anyway
- so the categories are the only thing left to display

I would prefer to see mockups or sketches, no matter how quick and dirty, before we rush into implementation. :) That way we can see how stuff works, and don't waste work on things which don't.

korelstar:
If you prefer to talk about mockups, then imagine that my implementation is a mockup ;-)

In order to make the list slim, we could remove the favorites item, of course. And note, that the search will be (re)moved, too (see #43). But we cannot remove the recent item, since otherwise, you cannot go back from a specific category to the whole list of notes. We can rename this item though.

However, the list of categories can be quite long (I'm using 11 first-level categories + uncategorized, in my personal notes). This would take much space from the screen (especially on mobile screens) and would require scrolling down in order to see the first notes items.

We would have to go with a second navigation then, like in the contacts app. In my personal opinion, this would be too distracting when writing notes. But if you prefer this, we could go this way.

I still think that my approach (showing categories on demand, using official Nextcloud design approaches) is best in giving access to categories and being not distracting when not using categories.

What's your proposal?

stefan-niedermann:
absolutely, looks great!

BloodyIron:
How do you set categories in the new method? Drag and drop, or right click on the note, or what?

BloodyIron:
Also, since this is the last item for the 2.5.0 milestone, any chance we could get this wrapped up and usable please? :DDD

korelstar:
See the screenshots in #228 for the current implementation of editing.

We have to do some testing and fixing the last bugs before releasing.

You can help in testing the latest state from master. Or you can increase the bounty on #8 in order to speed this up ;-)

PR #223, opened 2018-07-03
add setting for default file suffix

### LINKED ISSUE #24: Add setting to create new files with markdown file extension (.md)

korelstar:
fixes #24

- [x] add setting for default file suffix: drop-down box for selecting values (`txt`, `md`)
- [x] use setting when creating new files
- [x] use default value for empty settings


korelstar:
Ready for review

c33s:
please keep in mind, that i am **not** familar with the nextcloud code and plugin development at all, i don't know if all the things i am writing here are possible/not possibe/completly wrong/has sideeffects/are already possible with this code/...  please correct me if i am wrong.

as far as i can see you implemented a setting for the extension which can be changed in the plugin settings page. which is very nice, but isn't this only a per user setting?

in the issue #24 there was a lot of discussion about this topic:

- gui vs no gui
- default extension
- auto migration
- per user- vs global setting
- saving user from complicated things

often i read something like "saving the user from ..." but with this PR we don't save the user and leave the admin without an influence to that.

1. the simplest approach for this would have been, to declare it as **admin only feature**, so no user can change it, the default is defined by the admin in **`config.php`** (or wherever such a global setting is placed in nextcloud). so the admins can choose what they want to support.

2. decide which extension should be the default (i vote for `.md` here but i am ok with `.txt` as long this is **globally** changeable for **all** users (no no manual query in the database where i change it for all users and then define a stored procedure / trigger where i change it for all new created user)

3. for me after implementing and deciding step 1. & 2. the next step would have been to add it as **gui** = **per user setting** but maybe only with the option to **disable the user config** for certain groups or completly (we want to save the user from overcomplicated interfaces right? (it is not my way but some people expressed their concern about it. so its no sarkasm i really mean it)

with this PR it looks for me we are only a part having `[3]`. instead of `[1]` or `[1, 2]` or `[1, 2, 3]` where for me having `[1]` is the most important part.


what is really missing for me would to be able to configure the `fileSuffix` you added in `$defaults`, in a simply manner via cli (without browser, without login). for example add a simple string to `config.php` or run a cli cmd `php bin/console plugin:setting notes --fileSuffix='.md'` (borrowed the cmd syntax from symfony, i don't know if there is something equivalent in nextcloud)

service/notesservice.php
```diff
@@ -132,7 +132,7 @@ public function create ($userId) {
         // check new note exists already and we need to number it
         // pass -1 because no file has id -1 and that will ensure
         // to only return filenames that dont yet exist
-        $path = $this->generateFileName($folder, $title, "txt", -1);
+        $path = $this->generateFileName($folder, $title, $this->settings->get('fileSuffix'), -1);
```
c33s: why the change for `fileSuffix` instead of `extension` in all files?

service/settingsservice.php
```diff
@@ -17,8 +17,9 @@ class SettingsService
 	private $root;
 
 	/* Default values */
-	private $settings = [
+	private $defaults = [
 		"notesPath" => "Notes",
+		"fileSuffix" => ".txt",
```
c33s: please make this value globally changeable. this is the most important part of the issue your are fixing here.
maybe even configuring the `notesPath` globally would be nice.

service/settingsservice.php
```diff
@@ -33,16 +34,42 @@ public function __construct(
...
-		return $name ? (property_exists($settings, $name) ? $settings->{$name} : $default) : $settings;
+		if(is_object($settings)) {
+			// use default for empty settings
+			foreach($this->defaults as $name => $defaultValue) {
+				if(!property_exists($settings, $name) || empty($settings->{$name})) {
+					$settings->{$name} = $defaultValue;
+				}
+			}
+		} else {
+			$settings = (object)$this->defaults;
```
c33s: i personally would prefer to have a real settings class here and call the constructor with the defaults array. typecasting here makes me have a bad feeling but as i don't know the coding standard in this project and in nextcloud i assume you can ignore this. for projects where i approve the PR i would probably request a change.

in gerneral i would use symfonys option resolver for handling all the settings https://symfony.com/doc/current/components/options_resolver.html

service/notesservice.php
korelstar: Is your comment about the naming? I chose `fileSuffix` because the value includes the dot. I wanted to have the dot included, because it's more generic and I think that displaying `.md` to the user is more helpful than displaying `md`.

service/settingsservice.php
korelstar: We are not using symfony in this app, yet. Personally, I don't want to use another library just for this small thing. But please feel free to provide a PR with your proposed implementation.

service/settingsservice.php
korelstar: This PR is about a per user setting. I think that we should do a system-wide (admin) setting in a separate PR (if it is really needed).

korelstar:
This PR is about a per user setting. I think that we should do a system-wide (admin) setting in a separate PR (if it is really needed). What's your opinion on this, @jancborchardt ?

jancborchardt:
This is absolutely a per-user setting. It should be none of the admins concern what the default file extension for Notes is. Everyone might use different systems and workflows.

Also, the default should of course stay .txt while Windows still can’t really read .md by default. This is not a voting for which format we like better, but we need to stay compatible. :)

jancborchardt:
@camilasan @stefan-niedermann and @nextcloud/notes can you review this too? :)

korelstar:
@jancborchardt I implemented your change requests and I added the confirm buttons, you requested in nextcloud/server#1068.

grafik

Please review again ...

jancborchardt:
Good stuff, looks great! Thanks for also adding the confirm function! :)

korelstar:
> (We could also think about not having the confirm for the file extension and simply applying when you switch.)

Could we only think about this or should I remove the button? I have no opinion on this. You're the design boss. ;-)  
However, it would be nice to have a clear recommendation for the general usage of these buttons.

korelstar:
We can do this later, if needed.

jancborchardt:
> Could we only think about this or should I remove the button? I have no opinion on this.

Hehe – ok then I will say it would be great to not have to confirm but simply apply when the select value is changed. :)

PR #227, opened 2018-07-22
use new search API for Nextcloud >= 14

### LINKED ISSUE #43: Search is not where expected

korelstar:
Nextcloud 14 will introduce a complete new search, see nextcloud/server#9912. This PR implements the new search API if notes is running on Nextcloud >= 14. If notes is running on Nextcloud < 14, the old style search is used. This means, that this PR is fully backward-compatible. However, we can remove the old-style search when Nextcloud 13 isn't supported anymore.

Fixes #43.


korelstar:
I've also added two small, related things:
- [x] show loading icon in navigation while loading the list of notes (instead of "No notes found")
- [x] show "No search result for {{search}}" when using search (instead of "No notes found")

PR #228, opened 2018-07-24
Display resp. edit the current note's category

korelstar:
This is kind of a proof of concept in order to prepare a new functionality for changing a note's category (see #8). Now, we can go into a discussion about design aspects, @nextcloud/designers and @jancborchardt . I suggest the following design for displaying resp. editing the category of the current note.

### Default view for a uncategorized note
grafik

Clicking on the rename icon or on the category label =>

### Editing the category
The list is build from the existing categories.
grafik

Confirm the change or focus another element =>

### Changed category
grafik


stefan-niedermann:
Personally i would prefer using folders like in the mail app or the news app for consistency over introducing a new style inside the edit view. This would also keep the focus on the text and the organisational part at the left sidebar.

korelstar:
How do you change the current item's folder in those apps? Is this done over the navigation, only?

My approach was to display the folder view in the navigation in order to show notes from specific folders/categories (a filter like in other Nextcloud apps, see #210) but to provide changeability using the status bar (this #228). I divided this into two issues/PRs for better focusing on specific aspects. Maybe, this wasn't a good idea.

Please give more details on your favorite solution. The best would be to provide a mock-up :-)

Issue #204, opened 2018-06-05
SimpleMDE maintenance status


jancborchardt:
We use [SimpleMDE](https://simplemde.com/) as our base. Unfortunately looking at the repository at https://github.com/sparksuite/simplemde-markdown-editor it does not seem so maintained anymore.

As requirements and issues come in, maybe it makes sense to look for an alternative sooner or later. Maybe someone even took over maintainership in a fork or will do so soon.

No rush, but just to keep that in mind @nextcloud/notes. I guess @stefan-niedermann you have a similar situation with the Android library?

korelstar:
Looks like Markdown isn't that hip anymore. :-(

sunjam:
Here is [EasyMDE](https://github.com/Ionaru/easy-markdown-editor), a fork that has seen more maintenance.  Mentioned in [this thread](https://github.com/sparksuite/simplemde-markdown-editor/issues/685) on SimpleMDE.  I have not tested it.

juliushaertl:
I recently came across https://github.com/nhnent/tui.editor which looks pretty nice and seems to be well maintained (at least for now) :wink: I guess we should at least collect some options and compare them. Also in terms of possible WYSIWYG options which have been requested quite often and for stuff like #200

stefan-niedermann:
>  I guess @stefan-niedermann you have a similar situation with the Android library?

From time to time... At the moment, the Android-lib is experiencing much attention by its maintainer. Of course this can change every day, but currently i would **not** describe our lib as "not maintained".

jkaberg:
Moveing to something like previously suggested (https://github.com/nhnent/tui.editor) would be an HUGE improvement IMHO

jancborchardt:
@jkaberg it seems like you’re looking for the [Markdown app](https://apps.nextcloud.com/apps/files_markdown) which renders code on the left and rendered Markdown on the right.

We aim to merge the two as best as possible so you don’t need to have two separate views. Our first priority is notes. Then something we can look into for the future is stuff like iOS Notes and Google Keep does like lists, images etc.

jkaberg:
@jancborchardt actully no, I'm looking for a notes app with an WYSIWYG editor (as I have a hard time rembering syntaxes 😆), AFAIK neither the notes nor the markdown app has this

I've used https://stackedit.io/ in the past, and would love something like that with Nextcloud (I don't need the split pane thing in the markdown app)

korelstar:
Another hint from @biva in #214:

[...]

PR #258, opened 2018-10-02
Add proper fenced code-block formatting

tristandruyen:
Markdown defines tripple back-ticks (\`\`\`) as delimiter for code blocks, like
```
this
```
As i use notes extensively to save shell snippets, hack out readme's and the like, them being in a "normal" non monospaced font is extremely inconvenient...

It looks like my proposed change implements this in a quite minimal manner, and as far as my testing goes, doesn't break anything.

### DIFF
--- a/css/notes.css
+++ b/css/notes.css
@@ -244,7 +244,11 @@
 
 .CodeMirror .CodeMirror-code .cm-formatting-header:not(:first-child) {
     width: auto;
-    margin-left: 0px; 
+    margin-left: 0px;
+}
+
+.CodeMirror .CodeMirror-code .cm-comment {
+  font-family: MONOSPACE;
 }
 
 /* distraction free styles */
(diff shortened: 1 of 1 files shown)


tristandruyen:
BTW, While implementing this, I found it quite difficult to get into the simplemd & codemirror stuff, expecially as simplemd seems to be unmaintained since 2 years...
Would a PR to replace the markdown renderer with a more recent, extendable & actively maintained renderer be welcomed?

jkaberg:
@tristandruyen https://github.com/nextcloud/notes/issues/204 :smile:

korelstar:
Thank you very much for this fix!

Some help for enhancing the editor would be greatly appreciated. :-)
As @jkaberg wrote, there is already a discussion on alternatives to SimpleMDE (see #204). Firstly, we should decide for a good library, and then we can implement it in this app. You're welcome on participating in this discussion and, later, do the implementation :-)

PR #313, opened 2019-05-23
group list of notes by timeslots

korelstar:
Introduces labels for timeslots based on the last modified date, like we already have in the Android app and as discussed in https://github.com/stefan-niedermann/nextcloud-notes/issues/539

We have the following timeslots:
- Today
- Yesterday
- This week
- Last week
- This month
- Last month
- *[Month] [Year]*, e.g. *April 2019*
- *[Year]*, e.g. *2016*

*Month Year* is used for all dates which are before the last month but only in this year and last year. Older notes are grouped in *Year*.

Screenshot 1

Screenshot 2

Please review @nextcloud/notes @nextcloud/designers

### DIFF
--- a/css/app-navigation.scss
+++ b/css/app-navigation.scss
@@ -2,7 +2,7 @@
 	font-weight: bold;
 }
 
-#app-navigation > ul > li.app-navigation-caption {
+#app-navigation > ul > li.app-navigation-caption.caption-item {
 	padding: 0;
 	pointer-events: inherit;
 }
--- a/src/components/NavigationList.vue
+++ b/src/components/NavigationList.vue
@@ -97,6 +97,7 @@ export default {
 			return store.getters.numNotes()
 		},
 
+		// group notes by time ("All notes") or by category (if category chosen)
 		groupedNotes() {
 			if (this.category === null) {
 				return this.filteredNotes.reduce((g, note) => {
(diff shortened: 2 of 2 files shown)


jancborchardt:
@korelstar sorry for the late review! Looks very good, just 1 thing: Right now the headings are more important than the actual notes. Instead, these subheadings would be better in half-grey (`color: var(--color-text-maxcontrast);`) and not bold → like in the Mail app for the mail accounts.

korelstar:
Currently, it's just the default style for navigation labels from server. But I will change that ... Thanks :-)

jancborchardt:
@korelstar yes, we need to standardize that. :) cc @nextcloud/vue – but implementation (in several use-cases) comes before standardization. :) Thank you!

korelstar:
@jancborchardt I've changed the headings, but I think leaving them in bold and adjust the color only looks better:

| bold | normal |
|----|----|
| Screenshot | Screenshot |

What do you think?

skjnldsv:
A bit better I agree.
@korelstar when Jan approves, could you send your edit to the vue components? :)

jancborchardt:
@korelstar right, that’s also fine for now. :) (I’ll adjust other stuff at a later point and look at it again then.)

korelstar:
@skjnldsv The original style (`app-navigation-caption`) is not in `nc-vue` but in `server`. Should we change this in server? Or should we copy the style to `nc-vue` in order to be independent from `server`? Or should we have two style: one for "important" headings (like in category list) and one for "non-important" headings (like in this timeslot list)?

jancborchardt:
I would say let’s take the opportunity to create it in nextcloud-vue then? What do you think @skjnldsv

skjnldsv:
Ah you're right, we did not migrate the style yet ^^
@jancborchardt Yes that would be amazing! 

https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/AppNavigationItem/AppNavigationItem.vue
Also this would need a new AppNavigationCaption component (should be easy)

PR #323, opened 2019-06-16
print style

### LINKED ISSUE #179: CSS media query for printing style
At the moment the markdown format is lost when you're just left with txt files. It would be nice if there was an option to convert a Notes document into a PDF or directly call a window to send it to a printer. Thanks!

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56370322-css-media-query-for-printing-style?utm_campaign=plugin&utm_content=tracker%2F46751022&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751022&utm_medium=issues&utm_source=github).
</bountysource-plugin>

korelstar:
simple print media style, fixes #179

### DIFF
--- a/css/notes.scss
+++ b/css/notes.scss
@@ -1,1 +1,3 @@
 @import 'app-navigation.scss';
+@import 'print.scss';
+
--- a/css/notes.scss
+++ b/css/notes.scss
@@ -1,3 +1,2 @@
 @import 'app-navigation.scss';
 @import 'print.scss';
-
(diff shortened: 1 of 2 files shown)


jancborchardt:
Nice work! The parts of hiding `#header` & `#app-navigation` and focusing on the content could even be in core since they are useful for any app. Do you want to do a pull request to kick that off @korelstar?

cc @nextcloud/designers 
(We can also add further stuff like hiding the right sidebar etc. but it’s always good to start somewhere. :)

korelstar:
Yep, I can do that (later)

PR #348, opened 2019-07-26
update dependencies

korelstar:
webpack

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ php:
 addons:
   apt:
     packages:
-    - libxml2-utils
+      - libxml2-utils
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


korelstar:
Travis fails with error message
> ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)

Does anyone has an idea how to solve this?

This could be a temporary problem but also due to a change in a VM image, Nextcloud script or something else.

marcelklehr:
Travis has started progressively introducing Xenial to use for their builds [recently](https://changelog.travis-ci.com/xenial-as-the-default-build-environment-99476), which comes with a newer mysql version, perhaps you'll find out more in their changelog or googling in this context ;)

korelstar:
@marcelklehr 
Thanks for the hint! Looks like mysql is not started anymore by default. Fixed it by adding
```yml
services:
  - mysql
```

Issue #352, opened 2019-08-12
Default note view in preview mode


probastaller:
**Describe the solution you'd like**
I would like, if possible, to open the already written notes in preview formatted mode instead of edit mode. Or if it is not massive desirable, the option to chose default view in the settings would be great. In preview mode, links are clickable and text is cleaner and easier to read.

Thanks for your time :)

zaphodus:
It would be improve usability

frauhottelmann:
It woul increase the readability of the notes that already exist. The android app has a setting for this as well.

goofus:
Why is this closed? Does the option exist now?

goofus:
That's exactly the feature I am missing. Please add an option to open existing Notes in View Mode!

korelstar:
After realizing #331, there is no differentiation between edit mode and preview mode anymore - there will be only one single mode. Hence, this setting would just be relevant for a short period of time.

manfredu:
> After realizing #331, there is no differentiation between edit mode and preview mode anymore - there will be only one single mode. Hence, this setting would just be relevant for a short period of time.

This does makes sense to me. Is there an estimation when #331 will be realized?

korelstar:
No, sorry. We hadn't enough free time for this, yet.

I would accept a PR for a default view mode setting as interim solution, though.

Baptistou:
> After realizing #331, there is no differentiation between edit mode and preview mode anymore - there will be only one single mode. Hence, this setting would just be relevant for a short period of time.

There are still some differences between the two modes :
- Line endings are rendered in edit mode but preview mode requires 2 consecutive line endings to end a paragraph.
- Markdown Tables are rendered in preview mode but not in edit mode.
- Preview mode prevents you to modify a note accidentally.

Therefore, a default setting is still relevant.

arkadym:
> After realizing #331, there is no differentiation between edit mode and preview mode anymore - there will be only one single mode. Hence, this setting would just be relevant for a short period of time.

Yes, it is cool one, but still is not exaclty what we are discussing here, i still prefer able to see notes in preview mode by default.
Btw, #331 is missing tables support feature.

Fuseteam:
this would a nice setting, where should one look to add the settings? how does one test?

korelstar:
I think the following steps have to be done:

[...]

Issue #331, opened 2019-06-28
Integrate Nextcloud Text


korelstar:
Recently, Nextcloud Text was published: https://nextcloud.com/blog/nextcloud-introduces-collaborative-rich-text-editor/

This is a new rich text editor using markdown and providing collaborative editing. The plan is to integrate that editor into the notes app in order to have a common editor with all new features, but at the same time, make managing notes easy without thinking about files and directories (see discussion at https://github.com/nextcloud/text/issues/6)

@juliushaertl announced that he will start working on the integration of Text in the Notes app (https://github.com/nextcloud/text/issues/6#issuecomment-490775649). This issue here is just an information for users and developers that this integration is planned. I think it could fix #285, #38, #204, #310, #308, #329 and maybe some others. But let's see :wink:

korelstar:
Some information about the current state of this: I've implemented a proof-of-concept for the Text app's editor integration (see #652). Principally, it works already! However, there are some issues in the Text app that hinder me from finishing this. Mainly, these are:
- The Text app re-builds the whole Note when saving, using the CommonMark specification. This means that all formatting in the source code (e.g. line-breaks without effects for Markdown parsing) as well as additional non-Markdown parts is lost and Markdown is re-created using a predefined code style (e.g. list items are build with `* item`, even when they where originally `- item`). In my opinion, this is a show-stopper, since many people use notes with existing files that must still be processable by the original app that created those files. This issue is documented in https://github.com/nextcloud/text/issues/593 and https://github.com/nextcloud/text/issues/772 .
- Changing a note outside of the Text app (e.g. using the Notes Android app) leads to sync conflicts, even if the file wasn't changed in the Text app for some time. In such a situation, I expect the Text app to silently update the content. This issue is documented in https://github.com/nextcloud/text/issues/1212 .

Of course, there are some other things to be improved, since I've implemented a proof-of-concept, only. But these two are the biggest things and I think they will need much efforts.

Furthermore, I want to point to the different Markdown styles: The current edit mode allows for flexible line-breaks, e.g. split two list parts by inserting an empty line (see screenshots in #652). This is not possible in Text, since it's not (easily) possible in CommonMark. Honestly, I'm not happy with that, but maybe we have to get accustomed to this.

jotoeri:
Huh - just to give my 5 cents here:

[...]

PR #385, opened 2019-09-24
Default note view

zaphodus:
Set preview mode for notes that already exist.
#352

### DIFF
--- a/src/components/Note.vue
+++ b/src/components/Note.vue
@@ -130,7 +130,7 @@ export default {
 			store.commit('setSidebarOpen', false)
 			this.onUpdateTitle(this.title)
 			this.loading = true
-			this.preview = false
+			this.preview = true
 			NotesService.fetchNote(this.noteId)
 				.then((note) => {
 					if (note.errorMessage) {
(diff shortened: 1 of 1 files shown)


korelstar:
Thanks for your contribution!

Opening a note in edit mode is intended behavior since this app is primarily for editing notes and not just for viewing notes. However, I think you won't need this change anymore, after #331 is realized.

zaphodus:
Okay, that might be interesting.

manfredu:
> 
> Opening a note in edit mode is intended behavior since this app is primarily for editing notes and not just for viewing notes. However, I think you won't need this change anymore, after #331 is realized.

Then what is the proposed way to view notes? I use my notes as a searchable knowledge base and therefore I would very much like to have an option to see existing notes formatted and not in edit mode.

I do agree that the integration of Nextcloud Text would solve this issue (and others as well) . Has work on this already started?  When is this expected to be available?

Thanks!

manfredu:
Maybe this contribution could be merged as a temporary solution in case the realization of #331 will still take a while...?

korelstar:
No, sorry. We hadn't enough free time for this, yet.

I would accept a PR for a default view mode setting as interim solution, though.

PR #419, opened 2019-12-25
fix checkmark

### LINKED ISSUE #410: Checked Checkbox is too big
**Describe the bug**
The icon for a checked checkbox looks not good

**To Reproduce**
* make a list with checkboxes
```
[ ] Butter
[ ] 2 Kästen Mineralwasser
```

* click a checkbox
```
[x] Butter
[ ] 2 Kästen Mineralwasser
```

**Expected behavior**
The checkbox should look good

**Screenshots**
image

**Server (please complete the following information):**
- Notes app version: 3.0.3
- Nextcloud version: 17.0.1
- OS: Debian
- Web server:  Apache
- PHP version: 7.2
- Database: mySQL

**Client (please complete the following information):**
 - Browser (incl. version): Chrome 78.0.3904.97 and Microsoft Edge 44.18362.449.0
 - OS: Windows 10 v1909

korelstar:
fixes #410; uses checkbox design as used in tasks app (see https://github.com/nextcloud/tasks/pull/572)

## Before

| . | Nextcloud 15 | Nextcloud 18 beta 3 | 
|---|---|---|
| Windows 10 1909 | Screenshot | Screenshot |
| Debian 10 with Gnome  | Screenshot | Screenshot |

## After

| . | Nextcloud 15 | Nextcloud 18 beta 3 | 
|---|---|---|
| Windows 10 1909 | Screenshot | Screenshot |
| Debian 10 with Gnome  | Screenshot | Screenshot |

### DIFF
--- a/src/components/EditorEasyMDE.vue
+++ b/src/components/EditorEasyMDE.vue
@@ -167,22 +167,26 @@ export default {
 .CodeMirror .cm-formatting-task {
 	position: relative;
 	display: inline-block;
-	width: 1.8em;
-}
-
-.CodeMirror .cm-formatting-task.cm-meta::before {
-	content: '\2610';
+	width: 1.5em;
+	color: var(--color-main-background);
 }
 
 .CodeMirror .cm-formatting-task.cm-property::before {
-	content: '\2611';
+	background-image: var(--icon-checkmark-000);
 }
 
 .CodeMirror .cm-formatting-task.cm-meta::before,
 .CodeMirror .cm-formatting-task.cm-property::before {
-	font-size: 1.5em;
-	background: var(--color-main-background);
+	content: '';
+	width: 14px;
+	height: 14px;
 	position: absolute;
+	background-color: var(--color-main-background);
+	border: 1px solid #000;
+	border-radius: 1px;
+	background-position: center;
+	margin-top: 3px;
+	margin-left: 2px;
 }
 
 .CodeMirror .cm-formatting-task.cm-property + span {


jancborchardt:
Sorry I only saw this now – is there a specific reason we don't use the core checkbox style like in settings and elsewhere?

We should do the same in Tasks too cc @raimund-schluessler @nextcloud/design

korelstar:
The only difference is the color. The reason why tasks doesn't use the primary color is that the meaning of the checkmark is different:

- Settings: checkmark means active = primary color.
- Tasks and notes: checkmark means done = other color is better.

jancborchardt:
Ah ok, so the blue in the screenshot was just the emoji checkbox then? :)

And right, I remember we had the discussion in Tasks – makes sense and doesn't make the checkboxes so present. Ok! :)

korelstar:
Yep, the text "Checkbox: ☑ und ☐" is unicode only. The change is only the checkbox on the left side.

jancborchardt:
Hmmm, testing again and comparing also with e.g. iOS Notes it seems better to color it:
- it's more clear what is checked and what not
- motivates to check off unchecked items cause the colorless ones are boring
- we really could do with a bit more color in Nextcloud as stuff is quite bland :)

cc @nextcloud/designers and especially @ma12-co regarding Notes :)

korelstar:
Okay, In  #427, I changed the design to blue. It's now the design like in Nextcloud Text.

PR #422, opened 2019-12-28
add manifest.json

gko:
I use notes daily and love it. ❤️

I tried many options for a front end on mobile phone, but I find that the web version is just right (a part of some small things, that I will try to fix later).

## Motivation

The thing is that when you tap «Add to Home screen» in `chrome` it always saves the `Nextcloud` default app (i.e. `files`) which is in the way when you're trying to quickly add some note.

## Solution

I've looked through the code that is responsible of adding the manifest file and found that there is a default app -- `theming` that looks if the `apps/<app>/img/manifest.json` file is available and adds it otherwise it generates its own `manifest` in the [ThemingDefaults.php](https://github.com/nextcloud/server/blob/2442d78696313744bae8170c1b547f3b50ca8fcd/apps/theming/lib/ThemingDefaults.php#L347-L355) file:
```php
if ($image === 'manifest.json') {
	try {
		$appPath = $this->appManager->getAppPath($app);
		if (file_exists($appPath . '/img/manifest.json')) {
			return false;
		}
	} catch (AppPathNotFoundException $e) {}
	$route = $this->urlGenerator->linkToRoute('theming.Theming.getManifest');
}
```

but then the main [`.htaccess`](https://github.com/nextcloud/server/blob/2442d78696313744bae8170c1b547f3b50ca8fcd/lib/private/Setup.php#L525) file is redirecting all the files to `/`.

So I decided to make a [pull request](https://github.com/nextcloud/server/pull/18584) there where I would allow apps to have their proper `manifest` files.

## Result

With changes to server and the `manifest.json` within this PR I get:
image

And I'm able to save `Notes` app to desktop and access it quickly.

P.S. I also created a [`sandbox`](https://github.com/gko/nextcloud-sandbox) to be able to quickly debug apps.

### DIFF
--- a/img/manifest.json
+++ b/img/manifest.json
@@ -1,14 +1,14 @@
 {
 	"name": "Notes",
-	"start_url": "/apps/notes/",
+	"start_url": "../../../apps/notes",
 	"icons": [
 		{
-			"src": "/apps/notes/img/favicon-touch.png",
+			"src": "./favicon-touch.png",
 			"type": "image/png",
 			"sizes": "128x128"
 		},
 		{
-			"src": "/apps/notes/img/favicon-touch.svg",
+			"src": "./favicon-touch.svg",
 			"type": "image/svg+xml",
 			"sizes": "16x16"
 		}
--- a/img/manifest.json
+++ b/img/manifest.json
@@ -1,0 +1,17 @@
+{
+	"name": "Notes",
+	"start_url": "/apps/notes/",
+	"icons": [
+		{
+			"src": "/apps/notes/img/favicon-touch.png",
+			"type": "image/png",
+			"sizes": "128x128"
+		},
+		{
+			"src": "/apps/notes/img/favicon-touch.svg",
+			"type": "image/svg+xml",
+			"sizes": "16x16"
+		}
+	],
+	"display": "standalone"
+}


korelstar:
Thanks for your contribution! Your PR sounds reasonable. But since it depends on https://github.com/nextcloud/server/pull/18584, I'll wait for that PR to be merged until I merge this one.

However, I'm unsure about the paths in your manifest file: since these are not relative, I doubt that this works if Nextcloud is installed in a subdirectory. Could you please verify this?

gko:
@korelstar ok I fiddled a bit with different browsers. wherever I've installed `notes` app (in `custom_apps/` or in `apps/`) it's always resolved via `/apps/notes`.

korelstar:
@gko Any news on the pull request on server? I think they are waiting for you to optimize your code (see https://github.com/nextcloud/server/pull/18584#discussion_r361861634) ...

marcelklehr:
FYI, I implemented this with a slight hack in bookmarks, until server supports this natively: https://github.com/nextcloud/bookmarks/blob/master/lib/AugmentedTemplateResponse.php :shrug:

korelstar:
This has been finally implemented directly in server: https://github.com/nextcloud/server/pull/28459
However, thanks for your efforts!

PR #427, opened 2020-01-02
checkmark design

korelstar:
set checked design to blue, see https://github.com/nextcloud/notes/pull/419#issuecomment-569497851. Design is now equal to Nextcloud Text.

grafik


jancborchardt:
Good stuff, looks very nice! :)

PR #431, opened 2020-01-04
new undo design using notifications

### LINKED ISSUE #417: Design of Undo deletion
Follow up of #398.
The design has to be fixed, see: https://github.com/nextcloud/nextcloud-vue/issues/705

korelstar:
Fixes  #417, see discussion under https://github.com/nextcloud/nextcloud-vue/issues/705, especially https://github.com/nextcloud/nextcloud-vue/issues/705#issuecomment-549194181. Thanks to @paulschwoerer for help on implementing this. For now, I'm using the (old) `OC.Notification` API, since I want to be compatible to Nextcloud 15/16. We will switch to the new `OCP.Toast` API later.

grafik

@jancborchardt Is this how you imagined?

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,7 +32,9 @@ jobs:
     - name: Checkout
       uses: actions/checkout@master
     - name: Setup xmllint
-      run: sudo apt-get install --no-install-recommends -y libxml2-utils
+      run: |
+        sudo apt update
+        sudo apt install --no-install-recommends -y libxml2-utils
     - name: Setup xmllint problem matcher
       uses: korelstar/xmllint-problem-matcher@master
     - name: lint XML
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -48,6 +48,11 @@ return ['routes' => [
 		'verb' => 'POST',
 	],
 	[
+		'name' => 'notes#undo',
+		'url' => '/notes/undo',
+		'verb' => 'POST',
+	],
+	[
 		'name' => 'notes#update',
 		'url' => '/notes/{id}',
 		'verb' => 'PUT',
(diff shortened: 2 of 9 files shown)


jancborchardt:
Nice!

Let’s call it "Undo deletion", and does it maybe look more clickable if we make it a button?

Also, the `[ ]` were just a visualization for the button when I wrote the mockup in text – they should not be there. :)

korelstar:
I did a complete re-work of this (backend deletes directly and re-creates on undo, fronted uses @nextcloud/dialogs). Here are some impressions:

### deleting a single note

Screenshot

The Toast is closed after a timeout of 12 seconds. I used a longer timeout than normal due to the discussion under https://github.com/nextcloud/nextcloud-vue/issues/705#issuecomment-574156629

### deleting multiple notes

If another note is deleted before the timeout is expired, the Toast is adjusted and the timeout is reset:

Screenshot

Pressing "Undo delete" will restore all deleted notes within the timeout-period (here: three).

jancborchardt:
So @skjnldsv and I recently talked about Undo, so looping him in. :)

Also @korelstar only one thing about the notification: To prevent misclicks and accidental closing of the notification when you actually want to undo, can we remove the X button? It will automatically close anyway. :)

skjnldsv:
Oooh, never seen this!! :O
See https://github.com/nextcloud/nextcloud-dialogs/pull/106

korelstar:
Of course, we can remove the "x" button.

I'm already watching https://github.com/nextcloud/nextcloud-dialogs/pull/106, where a standardization is developed (there, the "x" button is removed). As far as I can see, that approach is similar, so I would switch so the standardized solution, when finished.

skjnldsv:
You rock!

PR #457, opened 2020-02-05
API: send HTTP 404 if note is not found

### LINKED ISSUE #454: Return appropriate HTTP statuscode on OCA\Notes\Service\NoteDoesNotExistException
Currently `HTTP 500` is returned from the server, in case that a client wants to `PUT` a locally modified note which does no more exist on the server.

Some potential candidates:
- `412 Precondition Failed` The server does not meet one of the preconditions that the requester put on the request header fields.
- `409 Conflict` Indicates that the request could not be processed because of conflict in the current state of the resource, such as an edit conflict between multiple simultaneous updates.
- `404 Not Found` The requested resource could not be found but may be available in the future. Subsequent requests by the client are permissible.
- `400 Bad Request` The server cannot or will not process the request due to an apparent client error

The candy on the ice cream would be a `410 Gone` if the file is in the trashbin.

korelstar:
fixes #454

### DIFF
--- a/lib/Controller/NotesApiController.php
+++ b/lib/Controller/NotesApiController.php
@@ -11,6 +11,7 @@ use OCP\IUserSession;
 use OCA\Notes\Service\NotesService;
 use OCA\Notes\Service\MetaService;
 use OCA\Notes\Service\InsufficientStorageException;
+use OCA\Notes\Service\NoteDoesNotExistException;
 use OCA\Notes\Db\Note;
 
 /**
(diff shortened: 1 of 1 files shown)


stefan-niedermann:
Fun thing is, it is already mentioned in the [API documentation](https://github.com/nextcloud/notes/wiki/Notes-0.2).

But the recently added `507`-header is missing. I just saw i have push rights, if you want i can add them to the docu.

korelstar:
That would be nice, thanks!

stefan-niedermann:
:heavy_check_mark: done

PR #491, opened 2020-04-13
API documentation for API v1

korelstar:
The API documentation is moved from the wiki to the main repository.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -18,36 +18,29 @@ We distinguish major and minor versions:
 - a major version comes with changes that are incompatible to the previous version and therefore would break old clients. Major versions come with a new base URL path.
 - a minor version has changes that are realized compatible to the previous version. Old clients can still use the current API endpoint, but they need adoption in order to use new features.
 
-Supported API versions can be queried using the [Nextcloud Capabilities API](https://docs.nextcloud.com/server/latest/developer_manual/client_apis/OCS/ocs-api-overview.html#capabilities-api).
+From Notes app version 3.3, supported API versions can be queried using the [Nextcloud Capabilities API](https://docs.nextcloud.com/server/latest/developer_manual/client_apis/OCS/ocs-api-overview.html#capabilities-api).
 
 A request like
 
 	curl -u test:test -X GET -H "OCS-APIRequest: true" -H "Accept: application/json" http://localhost/ocs/v2.php/cloud/capabilities 
 
-will return the following result:
+will return the following result (irrelevant attributes are omitted):
 
 ```json
 {
   "ocs": {
-    ...
     "data": {
-      ...
       "capabilities": {
-        ...
         "notes": {
-          "api_version": [
-            "0.2",
-            "1.0"
-          ]
-        },
-        ...
+          "api_version": [ "0.2", "1.0" ]
+        }
       }
     }
   }
 }
 ```
 
-The list of supported API versions is also provided in every response from the Notes API.
+From Notes app version 3.3, the list of supported API versions is also provided in every response from the Notes API.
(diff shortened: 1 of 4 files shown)


docs/api/README.md
```diff
@@ -47,7 +47,10 @@ It contains a coma-separated list of versions, e.g., `X-Notes-API-Versions: 0.2,
 
 ## Major API versions
 
-TODO Table
+| Version | Supported by app version | Remarks |
+|:-------:|--------------------------|---------|
+| **0.2** | since 1.0 (2015)         | *(deprecated)* |
+| **1**   | since 3.3 (April 2020)   | see [documentation](v1.md) |
```
stefan-niedermann: @korelstar Is it really `1` or `1.0`? I saw different spellings, it should be the same everywhere to avoid confusion.

korelstar: This table is about the major version, so it's `1`. (`0.2` is a special case, but this versioning approach is compatible to the "semver" approach [except that we don't have a patch version fro the API].) The major version determines the API's base path and all minor versions of the same major version are compatible to each other - if the points from the next section are realized on the client side.

Is this comprehensible? How could we improve the description?

docs/api/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+- when processing the JSON response, unknown fields must be ignored (e.g. if API version v1.0 does not define the note's attribute "tags", client must ignore such a field in order to be compatible with a possible future version v1.4 which defines such a field)
+- when processing the HTTP response code, a client must be able to handle newly introduced error codes (e.g. if API v1.0 does not explicitly define response code 5xx, the client must handle 
+
+In order to realize backwards compatibility between minor versions, a client must follow the following rules:
+
+- when sending a request which uses a feature that wasn't available from beginning of the used major version, the client has to cope with the situation that the server ignores parts of the request
+- when processing the JSON response, the server may ommit fields that where not available from beginning of the used major version
+
+If a client requires a certain feature, it should check the `X-Notes-API-Versions` HTTP header of the server response.
```
korelstar, line 67: Should we also send a special HTTP header for the full API version used in the current request/response? Example: a client requests `/index.php/apps/notes/api/v1/notes` and wants to know if a specific minor version is supported by the server (e.g. because that version introduced a new feature). Currently, the client has to check the `X-Notes-API-Versions` HTTP header from the response (which could be, e.g., `0.2, 1.1`), search for the major version `1` and look at the minor version.

In order to ease this check, we could add another `X-Notes-API-Version` (singular!) HTTP header which contains the currently used API version only, e.g. `1.1`.

@nextcloud/notes @stefan-niedermann

stefan-niedermann: I thought about it. I technically don't beed it, but it might be quiet useful for debugging and/or logging.

On the other hand: If we add many headers which aren't actually needed i am worrying about the traffic we are causing. We did so much work with the `ETags` and `If-Modified-Since`-stuff to drastically reduce the size of the responses and speed up everything and now we are adding many "long" headers which aren't necessarily needed.

Maybe a better option would be the possibility for the client to explicitely set a debug-header. If set, the server can add such kind of information (and maybe even set the server level for the context of the request to a more sensitive level)?

What do you think about this approach?

korelstar: I think the client should not be able to set the debug level - for security reasons. But we could check if server debug mode is active and send some debug information then. E.g, exact message and stack trace in case of an exception. What's about that?

stefan-niedermann: Sounds even better, but the stacktrace would have to be in the body, no? That might change the expection of the response...

I would start with "If server is in debug mode, attach the `X-Notes-API-Version` header as you proposed.

Seems to be an easy compromise.

PR #522, opened 2020-05-25
New icon for Notes app

jancborchardt:
Finally done with the new Notes icon long awaited by @stefan-niedermann ;) cc @korelstar 

This works well in the top bar with the other icons as well as in a circle on its own (like for the Android app):
new notes icon

The other favicon files are removed because theming should generate them automatically.


stefan-niedermann:
Android app:

grafik

jancborchardt:
Hmm, that’s strange. For me it is also weirdly displayed with a black border where there is none in the SVG, both Firefox and Chrome:
notes favicon
@juliushaertl @nextcloud/theming any idea?

jancborchardt:
So @juliushaertl said that it’s better to create these icons manually to ensure they are present even when the imagick PHP extension is missing. Will adjust the pull request accordingly. :)

juliushaertl:
So I also can reproduce the black border on my local setup, but honestly I have no idea why this is happening. For some reason imagick seems to assume a default stroke of `1` and `#000`. Adding `stroke-width="0" stroke="#ffffff"` to the svg path solves the border issue for me but still weired why this happens in the first place:

jancborchardt:
Created the favicons manually again now, please test. :)

juliushaertl:
Might still be worth to add the stroke workaround for themed instances that have this available :wink:

jancborchardt:
> Might still be worth to add the stroke workaround for themed instances that have this available wink

Hm, we don’t have this for any of the existing app icons though? So I’m not sure why it happens here and not for any other

jancborchardt:
This should be fine, as we don’t have it for any of the other apps either. Otherwise we should make sure it’s a general thing. :)

PR #555, opened 2020-06-27
debounce autotitle

### LINKED ISSUE #489: typing in the first line of a notes generates 2 activity entries for (almost) each character
**Describe the bug** (this is not really a bug, but anyways to stick to the template as much as possible)
When I type in the first line of a note (e.g. when starting a new note), on every character change the file gets saved (first activity entry) and the filename changes (second activity entry).

**To Reproduce**
1. type/make changes in the first line of a note (e.g. when starting a new one)
2. go to Nextcloud's Activity app/section
3. see a lot of entries for the just edited note (in fact two for (almost) every changed character)

**Expected behavior**
While it obviously makes sense to store the changes (on server side) as soon as possible to avoid data loss if e.g. the client disconnects due to network outage, those changes should only be written to disk once editing has "finished" (e.g. no changes for 5 seconds?). This way changes to the first line of a note can be done without "flooding" the activity stream.

**Server (please complete the following information):**
- Notes app version: 3.2.0
- Nextcloud version: 18.0.3

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/91106869-typing-in-the-first-line-of-a-notes-generates-2-activity-entries-for-almost-each-character?utm_campaign=plugin&utm_content=tracker%2F46751022&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751022&utm_medium=issues&utm_source=github).
</bountysource-plugin>

korelstar:
For new notes, the title is automatically changed. This PR debounces the title generation from saving the content. Therefore, a timer is set that generates a new title after ~~5 seconds~~ 2 seconds after the last typing in that note.

fixes #489

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -44,6 +44,12 @@ return ['routes' => [
 		'verb' => 'POST',
 	],
 	[
+		'name' => 'notes#autotitle',
+		'url' => '/notes/{id}/autotitle',
+		'verb' => 'PUT',
+		'requirements' => ['id' => '\d+'],
+	],
+	[
 		'name' => 'notes#update',
 		'url' => '/notes/{id}',
 		'verb' => 'PUT',
--- a/lib/Controller/NotesController.php
+++ b/lib/Controller/NotesController.php
@@ -191,14 +191,26 @@ class NotesController extends Controller {
 	/**
 	 * @NoAdminRequired
 	 */
-	public function update(int $id, string $content, bool $autotitle) : JSONResponse {
-		return $this->helper->handleErrorResponse(function () use ($id, $content, $autotitle) {
+	public function autotitle(int $id) : JSONResponse {
+		return $this->helper->handleErrorResponse(function () use ($id) {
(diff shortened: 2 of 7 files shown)


stefan-niedermann:
5 seconds is a little bit long, no? I think 1 second is by far enough?

stefan-niedermann:
Ah, i see, it's because of the activities issue...

Still i never created a new note which took longer than 5 seconds^^

korelstar:
I would not make this timeout lower than the auto save interval. Auto save interval is currently 2 seconds, but auto save will be executed also during typing - while auto title is only executed if you pause typing.

We could set auto save interval to 1 second and auto title interval to 2 seconds. That would still improve the situation of #489, I think.

What do you think of this?

/cc @violoncelloCH

stefan-niedermann:
Good enough for me :+1:

Issue #566, opened 2020-07-06
cannot create notes when using AWS S3 as main storage


lebuzzer:
there is error on notes creation : notes directory is not accessible: can't read file content.
application is broken after that.
only way to recover is to delete created file.

korelstar:
Please fill in the issue template (it's not a joke, we need that information in order to track down the error).

krakazyabra:
I have the same problem, but instead of AWS I'm using Minio

**Describe the bug**
Notes app is not working with S3 as primary storage

**To Reproduce**
Steps to reproduce the behavior:
1. Install fresh version of Nextcloud from docker nextcloud:18.0.7-fpm
2. Download and run minio/minio:RELEASE.2020-07-24T22-43-05Z-s390x
3. Set minio as default S3 primary storage
4. Try to use Notes app

**Expected behavior**
I can create, edit, manage Notes

**Screenshots**
![image](https://user-images.githubusercontent.com/26598203/88558121-6ed75b80-d02b-11ea-8b0b-4883cf7a121a.png)


**Server (please complete the following information):**
- Notes app version: 3.6.1
- Nextcloud version: 18.0.7
- OS: Debian buster
- Web server: nginx/1.19.1
- PHP version: PHP 7.3.20
- Database: percona/percona-xtradb-cluster:5.7.18

**Client (please complete the following information):**
 - Browser (incl. version): tested and confirmed on different devices
 - OS: tested and confirmed on different devices

**Log files**
Nothing connected to this problem

korelstar:
Please have a look again at the Nextcloud log. There should be an entry with an exception, there. The type of the exception is `Exception`. I need the exact message and stack trace in order to analyze this issue.

krakazyabra:
Only in debug mode I could find error messages. there are few of them, so here is the text of errors:  
<details><summary>A lot of logs</summary>
</p>

```
[
	{
		"reqId": "6SxMRCOxe4RaFcPAkYD7",
		"level": 2,
		"time": "2020-07-28T17:22:01+02:00",
		"remoteAddr": "10.28.36.165",
		"user": "company-auth-f74ba555-1d81-4cc5-a1a1-b1a0664e4ff4",
		"app": "core",
		"method": "GET",
		"url": "/apps/notes/",
		"message": "Failed to compile and/or save /var/www/html/apps/notes/css/notes.scss",
[...]

PR #586, opened 2020-08-17
Fix bug when using object storage (#566)

iolate:
blank files return false when using object storage as primary storage

* resolved: nextcloud#566

### DIFF
--- a/lib/Service/Note.php
+++ b/lib/Service/Note.php
@@ -38,7 +38,7 @@ class Note {
 	public function getContent() : string {
 		$content = $this->file->getContent();
 		// blank files return false when using object storage as primary storage
-		if ($content === false) $content = '';
+		if ($content === false && $this->file->getSize() === 0) $content = '';
 		if (!is_string($content)) {
 			throw new \Exception('Can\'t read file content.');
 		}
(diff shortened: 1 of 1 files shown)


korelstar (changes requested):
Thank you very much for your contribution! 🚀 

I'm afraid that your change breaks error handling. In the past, we had the problem that for some people a specific encrypted file can't be read (internal Nextcloud problem). In this case, `File:getContent()` also returns `false`. However, we must not return an empty content in this situation, but we have to throw an error (like before). Otherwise, it may happen that the user accidentally overwrites his note with empty content.

Therefore, we have to find another solution for this. I would be happy if you could help with finding a sustainable solution! 👍

iolate:
Then how about like this? I added a condition to check that file size is 0

korelstar:
Easy and effective - I think this is a good solution, thanks!

The linter complains that `Inline control structures are not allowed`. Could you please fix this? I will merge this PR then.

iolate:
Thank you!
Should I rebase my commits? I'm not familiar with pull request.. :(

korelstar:
Thank you very much! 🎉 

Everything is fine, now. :-)

PR #600, opened 2020-08-27
Unified search

### LINKED ISSUE #288: Search-filter for notes list doesn't use the notes' contents [$20]
Due to the speed optimization in #233, the content of the notes is not loaded anymore when loading the notes list. This is still a good decision, but as a consequence filtering notes using a keyword (resp. search string) doesn't "search" in the notes' contents - since filtering is performed on the client. Currently, filtering uses metadata such as the note's title, path, and category, only.

In order to (re-)introduce filtering by notes' contents, we have to implement a server-side search. Of course, the search should be performed asynchronously and has to cope with network errors, users who type faster than the network connection etc. (you know that stuff).

<bountysource-plugin>

---
There is a **[$20 open bounty](https://www.bountysource.com/issues/68334744-search-filter-for-notes-list-doesn-t-use-the-notes-contents?utm_campaign=plugin&utm_content=tracker%2F46751022&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751022&utm_medium=issues&utm_source=github).
</bountysource-plugin>

korelstar:
- [x] remove old search (not compatible to NC20)
- [x] add provider for Nextcloud's unified search (fixes #288)

## Screenshot
Screenshot Screenshot Dark Mode

## Open Questions
- [x] Which order index should we use? This is interesting, because notes will be shown by the `notes` provider and by the `files` provider. However, the `files` provider will open a note using the Text app and not using the Notes app. Therefore, it would be advantageous if notes will be shown first and files below. But I doubt that this is not desired.
- [x] Which icon should we use? For now, I use the app icon. Is this okay? Any other ideas?

@nextcloud/notes @nextcloud/designers


skjnldsv:
> * Which order index should we use? This is interesting, because notes will be shown by the `notes` provider and by the `files` provider. However, the `files` provider will open a note using the Text app and not using the Notes app. Therefore, it would be advantageous if notes will be shown first and files below. But I doubt that this is not desired.

Maybe 40 https://github.com/nextcloud/server/pull/22120#issue-463482520 ?

tnyeanderson:
I think the app icon looks good. Can we color-match it to the user's theme or is that a bad idea? If anything, just make it a dark gray because the dark black is a little jarring.

nickvergessen:
> I think the app icon looks good. Can we color-match it to the user's theme or is that a bad idea? If anything, just make it a dark gray because the dark black is a little jarring.

Well black is better as that inverts to white for dark-theme then. Also it's consistent with all other apps.

tnyeanderson:
I did not think about inverting for dark mode, that is a great point. Matching the user theme is probably not a good idea.

Looking through my app icons in the search results, I see .wav, .als, .html, .mp3, .vcf files are all a gray color. Maybe I'm missing something here. The gray can invert to a darker white as well, no? It's just easier on the eyes and looks much more clean. A stark black looks "outdated" if you know what I mean.

korelstar:
Thanks for you feedback, also in the Talk channel. I've implemented your suggestions and updated the screenshot. Time for a final review :smiley:

lib/AppInfo/SearchProvider.php
```diff
@@ -0,0 +1,83 @@
...
+		return $this->util->l10n->t('Notes');
+	}
+
+	public function getOrder(string $route, array $routeParameters): int {
+		if (strpos($route, 'files' . '.') === 0) {
+			return 25;
+		} elseif (strpos($route, Application::APP_ID . '.') === 0) {
+			return -1;
+		}
+		return 4;
```
skjnldsv, line 52: Default is far too high.
Do not inject yourself above default nextcloud apps please

lib/AppInfo/SearchProvider.php
```diff
@@ -0,0 +1,83 @@
...
+		return Application::APP_ID;
+	}
+
+	public function getName(): string {
+		return $this->util->l10n->t('Notes');
+	}
+
+	public function getOrder(string $route, array $routeParameters): int {
+		if (strpos($route, 'files' . '.') === 0) {
+			return 25;
```
skjnldsv, line 48: Why do you put yourself lower when you're on files?

lib/AppInfo/SearchProvider.php
skjnldsv, line 52:
```suggestion
		if (strpos($route, Application::APP_ID . '.') === 0) {
			return -1;
		}
		return 40;
```

skjnldsv (changes requested):

korelstar:
@skjnldsv This was attuned via https://cloud.nextcloud.com/call/b8rjbqp6 :

Joas Schilling @nickvergessen 28.08.2020 08:08:
> although notes sounds like a different case, because of how the info is opened. So maybe when you are on the files app you use something after files, when on any other app 4 and in notes -1

Jan C Borchardt @jancborchardt 09.09.2020 14:50:
> Late to the party as I was on vacation, but Joas Schilling feedback sounds good for me too :)

I asked if this is fine for everybody and nobody disagreed. But we can restart the discussion again, if you want :neutral_face:. The start was this:

> Which order index should we use? This is interesting, because notes will be shown by the `notes` provider and by the `files` provider. However, the `files` provider will open a note using the Text app and not using the Notes app. Therefore, it would be advantageous if notes will be shown first and files below. But I doubt that this is not desired.

What's your suggestion to deal with this problem? @skjnldsv

lib/AppInfo/SearchProvider.php
nickvergessen: So yeah, as korel already mentioned. Notes are files, but the notes app can present them better. So it should be the prefered app when you search anywhere over files, hence the `4` in general, because it should be before files which has 5. Then again on files you might prefer the files search (or not) which is why 25 was picked there, and -1 when in notes.

tnyeanderson:
@korelstar thank you for the changes to the icons... it looks really great!

For the result index... I definitely agree that our results need to be above the "Files" results. I think that putting a higher priority index is the best short term solution for that. This might be a more long term strategy (if it's not already implemented) but perhaps there should be a way to set one's application as the default for files based on type or location. So "All txt files in this folder should be listed to open with Notes"... so more granular than just overall indexes, because I think that's a recipe for confusion/disaster (as seen here, there are common cases where very "high" priority nextcloud apps need to be superceded by user apps... and I understand @skjnldsv hesitance because of this).

Obviously this would have to be implemented in nextcloud search itself, and would take a while to plan/implement if it ever is even considered, so we have to have a stopgap in the meantime.

jancborchardt:
Ah btw @korelstar same as for the Dashboard widget https://github.com/nextcloud/notes/pull/614#discussion_r491956877 – it would be cool if the subline would be the beginning of the note content instead of the category. :)

korelstar:
I implemented the last improvements (e.g. show note's excerpt instead of category @jancborchardt ) and I will merge this now, due to the pending major release.

@skjnldsv Sorry that I can't wait for your answer, but we let's try it with the current implementation and fix it afterwards, if there is a problem with it.

raimund-schluessler:
You can still filter notes by the unified search input, server emits an event when the search string is updated, see e.g. https://github.com/nextcloud/tasks/blob/master/src/main.js#L90-L96

korelstar:
@raimund-schluessler Thanks for the hint, I will have a look into that. But client-side filtering didn't work anymore for some time, since we do not load the mote's content for all notes anymore. Hence, searching/filtering must be realized on behalf of the server. Therefore, I think the current approach is fine, at least for now.

PR #614, opened 2020-09-20
Dashboard widget for NC20

### LINKED ISSUE #587: Notes widget for new dashboard
Nextcloud 20 introduces a new dashboard, see https://github.com/nextcloud/server/issues/20930

In the first step, the notes widget simply is a list of notes in the same order like in the notes app. A click on an item will open the notes app with the respective note.

Unfortunately, providing a dashboard widget requires to use the new bootstrap process (see https://github.com/nextcloud/server/issues/20573) which is incompatible to old Nextcloud server versions. Since I'm not able to maintain multiple notes app versions in parallel, this means that from version 3.7.0, the notes app will only support Nextcloud 20 (and above).

korelstar:
closes #587, see also https://github.com/nextcloud/server/issues/20930

| Empty content | with notes |
|---|---|
| Empty content | Widget |
| Darkmode (outdated!) Empty content Darkmode | Darkmode (outdated!) Widget Darkmode |

Everything fine? @nextcloud/designers @nextcloud/notes

### DIFF
--- a/css/global.scss
+++ b/css/global.scss
@@ -1,0 +1,1 @@
+@include icon-black-white('notes', 'notes', 1);
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -8,6 +8,7 @@ use OCP\AppFramework\App;
 use OCP\AppFramework\Bootstrap\IBootstrap;
 use OCP\AppFramework\Bootstrap\IBootContext;
 use OCP\AppFramework\Bootstrap\IRegistrationContext;
+use OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent;
 
 class Application extends App implements IBootstrap {
 	public const APP_ID = 'notes';
@@ -19,6 +20,11 @@ class Application extends App implements IBootstrap {
 
 	public function register(IRegistrationContext $context): void {
 		$context->registerCapability(Capabilities::class);
+		$context->registerDashboardWidget(DashboardWidget::class);
+		$context->registerEventListener(
+			BeforeTemplateRenderedEvent::class,
+			BeforeTemplateRenderedListener::class
+		);
 	}
 
 	public function boot(IBootContext $context): void {
--- a/src/components/Dashboard.vue
+++ b/src/components/Dashboard.vue
@@ -1,0 +1,121 @@
+<template>
+	<DashboardWidget :items="items"
+		:loading="loading"
+		:show-more-text="t('notes', 'notes')"
+		:show-more-url="showMoreUrl"
+	>
+		<template v-slot:default="{ item }">
+			<DashboardWidgetItem
+				:target-url="getItemTargetUrl(item)"
+				:main-text="item.title"
+				:sub-text="subtext(item)"
+			>
+				<template v-slot:avatar>
+					<div
+						class="note-item"
+						:class="{ 'note-item-favorite': item.favorite, 'note-item-no-favorites': !hasFavorites }"
+					/>
+				</template>
+			</DashboardWidgetItem>
+		</template>
+		<template v-slot:empty-content>
+			<EmptyContent icon="icon-notes">
+				<template #desc>
+					<p class="notes-empty-content-label">
+						{{ t('notes', 'No notes yet') }}
+					</p>
+					<p>
+						<a :href="createNoteUrl" class="button">{{ t('note', 'New note') }}</a>
+					</p>
+				</template>
+			</EmptyContent>
+		</template>
+	</DashboardWidget>
+</template>
+
+<script>
+import { DashboardWidget, DashboardWidgetItem } from '@nextcloud/vue-dashboard'
+import { EmptyContent } from '@nextcloud/vue'
+import { generateUrl } from '@nextcloud/router'
+import { getDashboardData, categoryLabel } from '../NotesService'
+
+export default {
+	name: 'Dashboard',
+
+	components: {
+		DashboardWidget,
+		DashboardWidgetItem,
+		EmptyContent,
+	},
+
+	data() {
+		return {
+			loading: true,
+			items: [],
+			hasMoreItems: false,
+		}
+	},
+
+	computed: {
+		showMoreUrl() {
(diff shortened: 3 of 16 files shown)


claell:
One thought: Does this look coherent with the other available widgets?

korelstar:
I'm using the [Nextcloud vue components for dashboard](https://github.com/nextcloud/nextcloud-vue-dashboard) (i.e. the DashboardWidgetItem), so I would say yes. But I don't have a complete overview over all other existing dashboard widgets, since not every app has released the dashboard yet and screenshots in respective PR's are often outdated. But that's why I'm asking for reviews, here! :smiley:

lib/AppInfo/Application.php
```diff
@@ -19,9 +19,11 @@ public function __construct(array $urlParams = []) {
 
 	public function register(IRegistrationContext $context): void {
 		$context->registerCapability(Capabilities::class);
+		$context->registerDashboardWidget(DashboardWidget::class);
 	}
 
 	public function boot(IBootContext $context): void {
+		\OCP\Util::addStyle('notes', 'global');
```
nickvergessen, line 26: Dashboard has an event which you can use instead of loading your css all the time (even on api requests you make it calculate paths)

css/global.scss
```diff
@@ -0,0 +1 @@
+@include icon-black-white('notes', 'notes', 1);
```
nickvergessen, line 1: This is a full request for just your icon? :( 
Use the dashboard template render event please

nickvergessen (changes requested):

lib/AppInfo/Application.php
korelstar: I need this style also for unified search. Is there also an event over there?

css/global.scss
korelstar: Baahm! But it's *my* icon! :laughing: 

To be honest: I just took this approach from the deck app, see
https://github.com/nextcloud/deck/blob/v1.1.0-beta2/lib/AppInfo/Application20.php#L87 and
https://github.com/nextcloud/deck/blob/v1.1.0-beta2/css/deck.scss#L1 - I think they should also be blamed  :sunglasses:

src/components/Dashboard.vue
```diff
@@ -0,0 +1,106 @@
...
+						:class="{ 'note-item-favorite': item.favorite }"
+					/>
+				</template>
+			</DashboardWidgetItem>
+		</template>
+		<template v-slot:empty-content>
+			<EmptyContent icon="icon-notes">
+				<template #desc>
+					<p class="notes-empty-content-label">
+						{{ t('notes', 'No notes found') }}
```
jancborchardt, line 26: Slightly better wording would be "No notes yet" since that communicates the proper state, instead of sounding a bit like they were lost. ;)

src/components/Dashboard.vue
```diff
@@ -0,0 +1,106 @@
...
+					</p>
+					<p>
+						<a :href="createNoteUrl" class="button">{{ t('note', 'Create note') }}</a>
```
jancborchardt, line 29: A bit better: "New note" to use the same wording as we do in the app itself for the button in the top left.

src/components/Dashboard.vue
```diff
@@ -0,0 +1,106 @@
...
+		getItemTargetUrl() {
+			return (note) => {
+				return generateUrl('/apps/notes/note/' + note.id)
+			}
+		},
+	},
+
+	methods: {
+		subtext(item) {
+			return categoryLabel(item.category)
```
jancborchardt, line 83: Instead of using the category for the subline, which is not _soo_ important (and can also return "No category" in lots of cases, which results in noise), could we use the beginning of the note instead (except the title of course)? :)

That would then be very similar to the Android app main view and give nice context to each note.

jancborchardt (changes requested):
Wow, awesome work @korelstar! :) Mostly small feedback inline.

src/components/Dashboard.vue
stefan-niedermann: Maybe `Add note`? I think in the Deck app it is called `Add card`, so maybe we can find a wording which fits to all those cases? (Maybe it's not that important though)

claell:
> I'm using the [Nextcloud vue components for dashboard](https://github.com/nextcloud/nextcloud-vue-dashboard) (i.e. the DashboardWidgetItem), so I would say yes. But I don't have a complete overview over all other existing dashboard widgets, since not every app has released the dashboard yet and screenshots in respective PR's are often outdated. But that's why I'm asking for reviews, here! 😃

I was only looking at the screenshots in the description of https://github.com/nextcloud/server/issues/20930. Those might be outdated, though.

css/global.scss
nickvergessen: Well, sorry if it sounded like blame, that was not the idea.
But you can use this event `OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent`

src/components/Dashboard.vue
jancborchardt: I’d say it’s a bit different in both cases. There’s also e.g. Mail where you wouldn’t say "Add mail" or "Add message". :) For cards the "Add card" fits, but for notes the "New note" wording is a bit more fitting. Does that make sense?

korelstar:
Thanks for your review, everybody!

Now, I've implemented:

- [x] use "New note" instead of "Create note"
- [x] use "No notes yet" instead of "No notes found"
- [x] use `BeforeTemplateRenderedEvent` for loading the stylesheet @nickvergessen 
- [x] show excerpt instead of category as subtext
- [x] some minor optimizations

@jancborchardt If there are no favorite notes, the widget looks like this:
Screenshot

jancborchardt (approved):
Tested and works perfectly, great work @korelstar! :)

korelstar:
@nickvergessen I think I've implemented your requests, so I'm merging this now, since I can't wait for the review anymore due to the pending release. Please give an info if there is something outstanding, I will fix this afterwards.

nickvergessen:
Yeah sorry all fine, didn't know you are still waiting for me (also I have 200+ unread notifications atm :( )

PR #630, opened 2020-11-22
fix(Notes Dashboard Widget): avoid HTTP 500 due mid-UTF8 excerpt truncation

bentolor:
Fix courtesty due to @D-side

This issue occurs when 100 bytes past the start of the note is inside
a UTF-8 multibyte character. E. g. if a note is comprised of 99
single-bytes and 100th multibyte.

This is very easy to hit in Russian, since it uses a Cyrillic charset
(2 bytes per UTF-8 char), but spaces and punctuation from ASCII
(1 byte each). But since the condition is so obscure, some texts
work and some don't seemingly at random (except not!).

Based on https://stackoverflow.com/q/9087502/ it seems that
using  `mb_substr` with `utf-8` seems to be the better approach

### DIFF
--- a/lib/Service/Note.php
+++ b/lib/Service/Note.php
@@ -64,12 +64,12 @@ class Note {
 		if (!empty($title)) {
 			$length = strlen($title);
 			if (strncasecmp($excerpt, $title, $length) === 0) {
-				$excerpt = substr($excerpt, $length);
+				$excerpt = mb_substr($excerpt, $length, null, "utf-8");
 			}
 		}
 		$excerpt = trim($excerpt);
 		if (strlen($excerpt) > $maxlen) {
-			$excerpt = substr($excerpt, 0, $maxlen) . '…';
+			$excerpt = mb_substr($excerpt, 0, $maxlen, "utf-8") . '…';
 		}
 		return str_replace("\n", "\u{2003}", $excerpt);
 	}
--- a/lib/Service/Note.php
+++ b/lib/Service/Note.php
@@ -62,13 +62,13 @@ class Note {
 		$excerpt = trim($this->noteUtil->stripMarkdown($this->getContent()));
 		$title = $this->getTitle();
 		if (!empty($title)) {
-			$length = strlen($title);
+			$length = mb_strlen($title, "utf-8");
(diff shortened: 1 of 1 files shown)


bentolor:
Quote a bunch of automated tests are failing but it seems due to a infrastructure problem, not my code change…

lib/Service/Note.php
```diff
@@ -60,12 +60,12 @@ public function getExcerpt(int $maxlen = 100) : string {
 		if (!empty($title)) {
 			$length = strlen($title);
```
korelstar, line 61: We have to switch from `strlen` to `mb_strlen`, otherwise we will have problems if the title contains multi-byte characters.

lib/Service/Note.php
```diff
@@ -60,12 +60,12 @@ public function getExcerpt(int $maxlen = 100) : string {
...
 			if (strncasecmp($excerpt, $title, $length) === 0) {
-				$excerpt = substr($excerpt, $length);
+				$excerpt = mb_substr($excerpt, $length, NULL, "utf-8");
 			}
 		}
 		$excerpt = trim($excerpt);
 		if (strlen($excerpt) > $maxlen) {
```
korelstar, line 67: We should use `mb_strlen` here, too.

korelstar (changes requested):
Thank you very much for your contribution! Although I wasn't able to reproduce the issue, it's very reasonable that this will fix it. Before I merge this, please

- replace `strlen` by `mb_strlen` in order to make sure that this won't break anything (see comments)
- rebase the PR, since I just fixed the failing CI tests (GitHub changed something recently)
- sign your work, see https://github.com/nextcloud/server/blob/master/.github/CONTRIBUTING.md#sign-your-work

lib/Service/Note.php
bentolor: Sounds obvious. Thanks for pointing out!

bentolor:
Thanks, @korelstar for your quick and helpful feedback.

I think I succeeded in adopting your proposed changes as well as mastering all the hurdles to get my MR ready for integration!

Thank you again for you quick feedback!

korelstar (approved):

Issue #200, opened 2018-06-04
Context-based formatting popover


jancborchardt:
To keep in line with our distraction-free design, but also to make the formatting easier to use, we should introduce a context-based formatting popover similar to what Medium.com has. Whenever you select text, the formatting options show:
![](https://cdn-images-1.medium.com/max/1600/1*GwPgFezPFHMKqN1EXBlhzg.png)

Then when you bold the text, it would insert the `**asterisks**` around the selected word, or for links it would `[insert the link tag and focus it on the url part](https://…)`.

cc @nextcloud/notes

exekutive:
What normal layperson would want to see a bunch of asterixes in their note?? How about implementing ACTUAL FORMATTING like bolt type and bullets so that notes written for personal use look nice and are actually READABLE. You complain that people try to write their own apps, rather than changing this one, but you refuse to change the scope. Markdown is for nerds only and guarantees that this app will never be widely adopted. You don't have to make it a word processor. Just basic stuff like bold, bullets, maybe italic, and 3 sizes. 

The name of the app is NOTES, so focus on that instead of making it a 1337 text editor.

putt1ck:
@exekutive you mean not text though. To support text formatting you have to have a save format that is more than text. Something that tells the UI to render this **word** in bold and that this:
*  should be a be a bullet point line.

Markdown is just a way of doing this. How it looks when you open it is down to the tool you use to open it. Even junior 1337 types can write markdown as they type, but few would normally consume a markdown file in "raw", just as they wouldn't an RTF or HTML one.

jancborchardt:
@exekutive it's all done in steps, and development only happens when people work on it. Your tone serves only to demotivate people, so please read comments again before you post.

exekutive:
@ putt1ck Store it as whatever you like, but 99% people of people aren't going to learn markdown just so they can take down notes. You are exactly right ... how it looks is down to the tool. This app IS a tool so why not make it user friendly??

@jancborchardt You are shutting down topics suggesting how to improve usability, based on your own narrow vision. THAT is demotivating.

putt1ck:
There will be a **WYSIWYG**. Just like on Github.

stefan-niedermann:
@exekutive great to hear, we all have the same target: enhancing the user experience of this app. If you want to start contributing, you can take one of the issues labled with "Good first issue" and simply provide a Pull Request.

[...]

PR #652, opened 2020-12-27
Integrate Nextcloud Text as editor for notes

korelstar:
Currently, this is just a proof-of-concept. See #331 for more information.

### Before: Edit mode
Screenshot

### Before: Preview mode
Screenshot

### After: WYSIWYG edit mode (replaces old edit mode and old preview mode)
Screenshot

## Open Tasks
regarding Notes app
- [ ] remember last opened note
- [ ] auto title for new notes

regarding Text app
- [ ] don't break original file, see https://github.com/nextcloud/text/issues/593 and https://github.com/nextcloud/text/issues/772
- [ ] improve parallel changes from outside of app, see https://github.com/nextcloud/text/issues/1212
- [ ] use correct / other list/task syntax, see https://github.com/nextcloud/text/issues/1297

regarding Text integration in Notes app
- [ ] loading another note with same component (proof-of-concept uses a "hack" which does not allow for a loading indicator)
- [ ] "Link file" feature of the Text app could be replaced by a "Link note" feature (?)
- [ ] how to provide integrated images to 3rd-party apps using the Notes API?

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -6,10 +6,13 @@ namespace OCA\Notes\Controller;
 
 use OCA\Notes\Service\NotesService;
 
+use OCA\Viewer\Event\LoadViewer;
+
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\AppFramework\Http\ContentSecurityPolicy;
 use OCP\AppFramework\Http\RedirectResponse;
+use OCP\EventDispatcher\IEventDispatcher;
 use OCP\IRequest;
 use OCP\IURLGenerator;
 use OCP\IUserSession;
@@ -19,6 +22,8 @@ class PageController extends Controller {
 	private $notesService;
 	/** @var IUserSession */
 	private $userSession;
+	/** @var IEventDispatcher */
+	private $eventDispatcher;
 	/** @IURLGenerator */
 	private $urlGenerator;
 
@@ -27,11 +32,13 @@ class PageController extends Controller {
 		IRequest $request,
 		NotesService $notesService,
 		IUserSession $userSession,
+		IEventDispatcher $eventDispatcher,
 		IURLGenerator $urlGenerator
 	) {
 		parent::__construct($AppName, $request);
 		$this->notesService = $notesService;
 		$this->userSession = $userSession;
+		$this->eventDispatcher = $eventDispatcher;
 		$this->urlGenerator = $urlGenerator;
 	}
 
@@ -41,6 +48,7 @@ class PageController extends Controller {
 	 * @NoCSRFRequired
 	 */
 	public function index() : TemplateResponse {
+		$this->eventDispatcher->dispatch(LoadViewer::class, new LoadViewer());
 		$devMode = !is_file(dirname(__FILE__).'/../../js/notes-main.js');
 		$response = new TemplateResponse(
 			$this->appName,
--- a/src/components/Note.vue
+++ b/src/components/Note.vue
@@ -1,16 +1,21 @@
 <template>
-	<AppContent :class="{ loading: loading || isManualSave, 'icon-error': !loading && (!note || note.error), 'sidebar-open': sidebarOpen }">
+	<AppContent :class="{ loading: loading, 'icon-error': !loading && (!note || note.error), 'sidebar-open': sidebarOpen }">
 		<div v-if="!loading && note && !note.error && !note.deleting"
 			id="note-container"
 			class="note-container"
 			:class="{ fullscreen: fullscreen }"
 		>
 			<div class="note-editor">
-				<div v-show="!note.content" class="placeholder">
-					{{ preview ? t('notes', 'Empty note') : t('notes', 'Write …') }}
-				</div>
-				<ThePreview v-if="preview" :value="note.content" />
-				<TheEditor v-else :value="note.content" @input="onEdit" />
+				<component :is="viewer.component"
+					ref="texteditor"
+					:fileid="fileId"
+					:basename="title"
+					:active="true"
+					:has-preview="true"
+					mime="text/markdown"
+					class="text-editor"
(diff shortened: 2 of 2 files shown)


stefan-niedermann:
Great to see some progress here! :partying_face: 

Will you consider implementing [Context based formatting](https://github.com/nextcloud/notes/issues/200) also in this PR? Because if you show the toolbar permanently and users will get used to it, it will be hard to change the behavior *again* in the future.

korelstar:
I like context based formatting and I would like to see it in this app. But with realizing the Text integration into the Notes app, there will be no code left in the Notes app regarding editor features. This means, that if the Notes app should provide context-based formatting, then this has to be implemented in the Text app. Therefore: no, this will be not part of this PR.

Nevertheless, context-based formatting could be classified as a release-critical feature. This would mean, that we won't merge this PR until Text has implemented this feature. As listed in https://github.com/nextcloud/notes/issues/331#issuecomment-751512413, there are also other issues of this criticality. Hence, the time for merging this seems to be far away anyway.

On the other hand, I don't even know if context-based formatting is a desired feature in the Text app. This has to be discussed with the Text people and the design people. Maybe you could assess this from your perspective? @nextcloud/text @nextcloud/designers (and also /cc @nextcloud/notes )

juliushaertl:
Nice :)

Let me comment on some of the points here:

> regarding Text app
> don't break original file, see nextcloud/text#593 and nextcloud/text#772
> use correct / other list/task syntax, see nextcloud/text#1297

I feat that we might never be able to fully preserve the original formatting. The main reason is that the WYSIWYG editor does not preserve the original format but converts the markdown document to a different document format (a json tree with different node types like h1/li and annotation types like italic/bold) and then on save converts this json format back to markdown. Taking lists as an example we just don't know when saving if the original markdown file was written with `*` or `-` as list indicators. This was decided to be an acceptable downside of the approach since we mainly aimed for a rich content editor here rather than a markdown editor. However we should of course preserve as much as possible so the open enhancement requests about frontmatter or tables for example should definitely be implemented at some point.

> improve parallel changes from outside of app, see nextcloud/text#1212

Yes, we have some ideas about automatically merging the outside changes into the history of the collaboratively edited document, however this will be quite some effort to get to.

> On the other hand, I don't even know if context-based formatting is a desired feature in the Text app. This has to be discussed with the Text people and the design people. Maybe you could assess this from your perspective?

I'm open to discuss this, the main personal concern here is that the formatting with a popover like shown by @jancborchardt in https://github.com/nextcloud/notes/issues/200 is not really discoverable in my oppinion.

Regarding the general usage of the text app component I think your approach is fine for a first attempt. Ideally we would have the editor available as a dedicated libary that could then just be used in apps. I started in https://github.com/juliushaertl/nextcloud-text/ however didn't have much time to continue with that recently.

Also let me know if I shall add you to the text conversation on our Nextcloud Talk instance, I think you already have an account there if you want to discuss something quickly.

juliushaertl:
It might also be worth to have a look at https://gitlab.com/collectivecloud/collectives which @azul and @mejo- built last year as a prototype fund project which is based on Nextcloud Text in a similar way.

korelstar:
Hey Julius,

thanks for your clear statement on the WYSIWYG problematic. This helps me to look at this from the right perspective. I'm pretty sure, that some users won't like this approach, since they use the Notes app with externally generated files. But you're right, we have to weigh the pros and cons.

The most important thing IMHO is that the Notes app should still work seamlessly with the Android Notes app. My first impression was, that there will some more issues after integrating Text (besides the synchronization issue) due to the strong CommonMark based WYSIWYG approach. I will investigate that further (maybe @stefan-niedermann can also do some tests?).

Nevertheless, please add me to the Talk channel for Text, I'm already on that NC instance for other channels.

I also took a look on the collectivecloud project (accidentally) and took some code from their integration approach. That was a good starting point! :-)

azul:
Hey @korelstar ... i thought that approach looked familiar 😃 - happy it was useful.

It has worked for us so far but it's really not a clean implementation. I'll be back to working more on the collectives app in march and maybe i can also take a stab at a cleaner way of using the text app in other contexts.

inthreedee:
> regarding Text app
> don't break original file, see nextcloud/text#593 and nextcloud/text#772
> 
> ...This was decided to be an acceptable downside of the approach since we mainly aimed for a rich content editor here rather than a markdown editor.

When deciding how to handle this situation, please see the comments in the two issue threads mentioned.  In particular, https://github.com/nextcloud/text/issues/772#issuecomment-653942643 seems to resonate.  I agree that the tradeoffs as discussed by @juliushaertl are reasonable for a rich content editor but not for one that is set up as the default editor of (and then destroyer of) markdown files.  For example, on iOS, it seems to be the only option now; the old edit in plain text option seems unavailable even when the new Text app is disabled.

If this new editor is not intended to perform the functions of a proper markdown editor, then it probably should not handle markdown files at all given the destructive side-effects.  Give these "Nextcloud rich text format" files and notes that it creates their own file extension to avoid this situation where markdown files created in other applications get destroyed.

Basically, if it is only capable of handling markdown files in the "Nextcloud rich text format", it shouldn't be using .md files because nobody else uses the "Nextcloud rich text format" for their .md files.  The best time to make this kind of change from a user's perspective would be during a large integration like this.

vincowl:
Hi there,
Could someone please tell us wether this feature has a chance to be implemented one day or if after a whole year without any activity, we should consider this as an implicit "forget it" answer ? I personnaly would love to use Notes, but having 2 markdown editors hosted on nextcloud (and one is the nextcloud's ref) is a strong no go for me. I would love to write my md files online or on my mobile, but would love far more something that can be cleanly integrated in the other modules (can be edited / formatted the same way). And I think this should imply choosing a default markdown parser and editor. So why not Text ?

juliushaertl:
We are going to expose the text app editor through an API, so I'll close this in favour of https://github.com/nextcloud/notes/pull/958

PR #665, opened 2021-01-13
fix long duplicate title generation

skazi0:
Long titles were trimmed to 100 characters even after appending numbered suffixes for duplicates. This caused infinite recursion / hanging of the service.
This fix removes some more characters from the end of title if there is not enough room for the suffix.

Real life example for this problem is when user notes long URLs (e.g. google search results) in separate notes. First 100 chars can often be the same and this triggers the duplicate handling.


korelstar (approved):
Nice! Thank you very much for your contribution!

Issue #627, opened 2020-11-01
Sync API timing problems


davidbannon:
This relates to #621 but is much more specific and probably better presented as a stand alone issue.
I am building a sync between tomboy-ng and Nextcloud Notes. The Tomboy standard relies on meta data stored in xml and I am finding difficulties working with NextCloud which does not.

Unless there is a way around this issue (that I cannot see) it is probably also causing problems with other applications using the API. 

Because the 'modified' time returned via the API does not respect things like changes to a note's name, I must use the pruneBefore parameter to a GET to determine what notes have changed on NextCloud.  So, the process is to request a list of notes from NextCloud and record its time stamp (LastSyncTime). We work out what needs to happen and download and upload as necessary. That takes some time so we end up with uploaded notes getting NextCloud time stamps later than the recorded LastSyncTime.  

If no changes are made at either end and another sync is run, the uploaded notes (remember they were uploaded after the LastSyncTime) are seen as new or changed and will be downloaded. Putting aside the network traffic issue, they will then get a wrong LastChangeDate and will trigger a false 'clash' if the user does make a change locally between those syncs. Confusion and potential data loss.

A similar problem happens if new/changed notes are uploaded **before** taking a pruneBefore GET. 

Is there a way to avoid this that I have not spotted ?  It seems to me that the difficulties will be with us unless NextCloud Notes finds a way to maintain the necessary meta data outside of the note files them selves.

NextCloud seems to be maintaining two time stamps, the one associated with 'modified' and the one that is used by pruneBefore. I'm guessing that pruneBefore uses the file's file system date stamp, just where does 'modified' time come from ?   Personally, I believe that 'modified' should mean any modification including the note name !

Davo

davidbannon:
Another approach would be to allow a means to set the (pruneBefore) time on a note when its uploaded. Settings it's 'modified' time does not help by the way.

If I am right that pruneBefore uses the file system datetime for the note file itself, then allowing it to be set would be hard but not impossible (given that I have no idea what interaction is allowed by Nextcloud to filesystem). 

It would be nice to see a bit of documentation about these two time stamps and how they are used.

Davo

korelstar:
Hey @davidbannon, I'm very sorry for my late reply. Unfortunately, I am very busy since some months.

I understand that a real last-modified time (which includes all metadata) would be nice, but unfortunately the Nextcloud architecture doesn't provide such a time and, however, I don't think it is necessary. Let's have a look behind the scenes:

[...]

Issue #621, opened 2020-10-13
Modified date/time is not updated when a note is renamed.


davidbannon:
Hi folks, love your work.
I have a beta (and only just beta) sync working with tomboy-ng ( https://github.com/tomboy-notes/tomboy-ng ) and I have found that when a note is renamed in nextCloud, its modified time (as returned under the sync api) is not updated. This means that the remote system is unaware of a change to the note and does not choose to download it.

Reproducing this bug will depend very much on how you personally choose to tickle the api .... 

Davo

stefan-niedermann:
Could you verify that the `ETag` changed?

davidbannon:
I don't think I have access to an ETag via the api do I ?  If not, I possibly cannot verify it.
Can you tell me what is / how to see the ETag ?
I can use a local, multipass based NextCloud server if necessary.
Davo

stefan-niedermann:
ETags are used to verify that something has changed on the server side. You can imagine it as an hash over the http response. you send the hash in your next request as value of an `If-None-Match`-Header and the server will return `304` HTTP status code if the response is the exact same as last time. If *anything* changed (e.g. the title), you will get a complete response.

See the [Notes API documentation](https://github.com/nextcloud/notes/blob/master/docs/api/v1.md#200-ok) and [more information about ETags](https://en.m.wikipedia.org/wiki/HTTP_ETag).

This is better as relying on the modified date, because in some situations one doesn't *want* to change the modified date. For example this would cause a order change when one ordered the notes by modified date, but this isn't actually wanted behavior.

Other scenarios might be "marking a note as favorite" - i wouldn't expect an updated modified date here, as it would cause a sort order change.

davidbannon:
OK, I reckon I can modify my code to read the etag but will need a bit of time to do so. But that really won't help. It won't tell me which note/notes has changed.   We need to have something that is a clear indicator that a particular server note is in need of being synced, if its not modified time, then what is it ?

If i am to rely on the eTag, and I get a non-match, I would need to download all the notes and compare them with local notes. Given that I have users with tens of thousands of notes, I really don't think thats a viable approach.

blue sky mode on :
If I could get a note by note eTag, when requesting a single note, that would help but would still be quite slow, too slow I expect. Better if you could add a hash of each note to the list of note attributes, i could save that locally and compare at sync time. It would also solve a very nasty problem where someone 're-establishes' a sync and has previously synced notes at both ends.   

Davo

[...]

PR #692, opened 2021-03-12
Prevent lost updates by using ETags for notes

### LINKED ISSUE #56: Manage concurrent edits of notes

korelstar:
Finally, I've implemented ETags for notes / If-Match HTTP header for the notes API. This fixes #56.

Includes:
- [x] API: expose the note's ETag and check note's ETag on update request (HTTP If-Match / 412) => new minor API version 1.2
- [x] documentation for API changes
- [x] Test cases
- [x] web-frontend: check for update conflicts and provide solutions

@stefan-niedermann @phedlund : Please have a look if you see any problems with these changes. They are downwards-compatible, so you don't need to change your implementation. However, it is strongly recommended in order to prevent lost updates. If you implement it, your implementation is still compatible with older (server) Notes app versions, since they just ignore the `If-Match` header.

### DIFF
--- a/docs/api/v1.md
+++ b/docs/api/v1.md
@@ -91,9 +91,11 @@ No valid authentication credentials supplied.
 | Parameter | Type | Description |
 |:------|:-----|:-----|
 | `id` | integer, required (path) | ID of the note to query. |
+| `If-None-Match` | HTTP header, optional | Use this in order to reduce transferred data size (see [HTTP ETag](https://en.wikipedia.org/wiki/HTTP_ETag)). You should use the value from the note's attribute `etag` or from the last request's HTTP response header `ETag`. | 1.2 |
 
 #### Response
 ##### 200 OK
+- **HTTP Header**: `ETag` (see [HTTP ETag](https://en.wikipedia.org/wiki/HTTP_ETag)). The value is identical to the note's attribute `etag` (see section [Note attributes](#note-attributes)).
 - **Body**: note (see section [Note attributes](#note-attributes)), example:
 ```js
 {
--- a/docs/api/v1.md
+++ b/docs/api/v1.md
@@ -20,7 +20,7 @@ The app and the API is mainly about notes. So, let's have a look about the attri
 | Attribute | Type | Description | since API version |
 |:----------|:-----|:-------------------------|:-------------------|
 | `id` | integer (read‑only) | Every note has a unique identifier which is created by the server. It can be used to query and update a specific note. | 1.0 |
-| `etag` | string (read‑only) | The note's entity tag (ETag) indicates if a note's attribute has changed. I.e., if the note changes, the ETag changes, too. Clients can use the ETag for detecting if the local note has to be updated from server and for [optimistic concurrency control](https://en.wikipedia.org/wiki/Optimistic_concurrency_control). | 1.2 |
+| `etag` | string (read‑only) | The note's entity tag (ETag) indicates if a note's attribute has changed. I.e., if the note changes, the ETag changes, too. Clients can use the ETag for detecting if the local note has to be updated from server and for optimistic concurrency control (see section [Preventing lost updates and conflict solution](#preventing-lost-updates-and-conflict-solution)). | 1.2 |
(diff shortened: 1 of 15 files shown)


korelstar:
Any feedback on this? Any plans to implement this in your apps @stefan-niedermann @phedlund?

Also cc @davidbannon since it is related to #621 and #627.

stefan-niedermann:
Yes, i would like to implement support for it, though i won't be able to make it in the next few weeks due to some other responsibilities and tasks i have to do. I also will probably need some help for the concrete implementation.

korelstar:
Okay that's fine. I think it would be good to finish your Room migration before starting with this, since this may require some DB changes. Of course, I can give some advises/help then. :-)

PR #694, opened 2021-03-12
new Settings API

korelstar:
Provides access to the app settings through the API. This was requested in https://github.com/stefan-niedermann/nextcloud-notes/issues/916#issuecomment-711148998. However, my implementation differs to the proposed API. Please check, if this is fine for you, too @stefan-niedermann . My approach was easier :grin: 
Please note, that the API does not provide the list of allowed file suffixes. This is static for now. We could add a dynamic list later. Is this fine?

- [x] new API endpoints for getting and setting app settings
- [x] documentation for API changes
- [x] Test cases

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ app_name=notes
 project_dir=$(CURDIR)/../$(app_name)
 build_dir=$(CURDIR)/build/artifacts
 cert_dir=$(HOME)/.nextcloud/certificates
+php_dirs=appinfo/ lib/ tests/api/
 
 
 all: dev-setup build
@@ -99,7 +100,7 @@ lint-phpfast: lint-php-lint lint-php-ncversion lint-php-cs-fixer lint-php-phpcs
 
 lint-php-lint:
 	# Check PHP syntax errors
-	@! find lib/ -name "*.php" | xargs -I{} php -l '{}' | grep -v "No syntax errors detected"
+	@! find $(php_dirs) -name "*.php" | xargs -I{} php -l '{}' | grep -v "No syntax errors detected"
 
 lint-php-ncversion:
 	# Check min-version consistency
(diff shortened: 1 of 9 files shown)


stefan-niedermann:
Now that looks awesome, i like the approach 🙂 

I am missing two things here in the documentation:
- Error codes, for example:
  - providing a invalid value as file extension
  - use `../` or `/` or `1337` as value for the folder
- Some kind of "reset" - A client does not know the localized term of "Notes" in the server app and has to guess therefore if the user wants to reset the settings to default

korelstar:
Thanks for the feedback!

Currently, the request succeeds even if the values are invalid. The server performs a sanitation of the values and sends back the fixed values.
- `fileSuffix`: if the value is not allowed, the default value is used.
- `notesPath`: the path is sanitized, i.e. unnecessary `./`, `X/../`, `//` are removed.

I've added the following sentences to the documentation:
> Empty values are replaced by the settings attribute's default value. All values are sanitized (e.g. prevent path traversal attacks, check allowed suffixes), so the result can differ from the request (the request will still succeed). The client may show an information to the user if the response differs from the request.

Is this sufficient?

stefan-niedermann:
Yes :)

So in my understanding, if i push the object:

```json
{
  "fileSuffix": null,
  "notesPath": null
}
```

the sanitizer will map it to roughly this values (and return it in the response):

```json
{
  "fileSuffix": ".txt",
  "notesPath": "Notes"
}
```

korelstar:
Yep, that's correct

stefan-niedermann:
I have kind of a [working implementation](https://github.com/stefan-niedermann/nextcloud-notes/pull/1091/files), i think. Looking forward to test this with a real server :)

korelstar:
> I have kind of a [working implementation](https://github.com/stefan-niedermann/nextcloud-notes/pull/1091/files), i think. Looking forward to test this with a real server :)

What do you mean with "real server"? You can checkout this branch and install it. Did you do such tests in the meanwhile?

stefan-niedermann:
Nah, nevermind :smile: I just don't have a test instance right now where i could push the state from the branch. When i have time i will set up one.

PR #711, opened 2021-05-02
New attribute "readonly" for read-only notes

### LINKED ISSUE #564: Handle read only files correctly
When opening and editing a `-rwxrwxrwx` file, everything is fine.
When opening a `-rwxr-xr-x `(owned by a user other than www-data), everything is fine.
But when editing a `-rwxr-xr-x `(owned by a user other than www-data), you can input the first few words(which doesn't actually get inputted into the file). Then you can not see any content but `File could not be loaded. Please check your internet connection.[Retry]`

Maybe check if a file is writable before opening a file? Because I have some files that are too important to lose. I can't risk making it 777.

korelstar:
- Checks if a note is writable
- sets the new attribute "readonly"
- API: announce this attribute over the API
- API: send 403 if a note can't be changed due to a read-only note
- web GUI: hides some buttons for read-only notes
- web GUI: disables editing for read-only notes

Closes #564

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,6 +66,7 @@ jobs:
         mkdir -p server/data/quotatest/files/
         cp -r notes/tests/reference-notes server/data/test/files/Notes
         cp -r notes/tests/reference-notes server/data/quotatest/files/Notes
+        chmod 444 server/data/test/files/Notes/ReadOnly/ReadOnly-Note.txt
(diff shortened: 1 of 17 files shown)


korelstar:
Are you fine with this @nextcloud/notes (especially @stefan-niedermann @phedlund)?

-----

`test-api (max)` is failing due to https://github.com/nextcloud/server/pull/26572 .

PR #747, opened 2021-08-07
add setting for view mode

### LINKED ISSUE #352: Default note view in preview mode

korelstar:
Adds a new setting "Display mode for notes" with options "Open in edit mode" and "Open in preview mode". If you open a note, it will be shown either in edit mode or in preview mode - depending on this setting. In both cases, you can still change the mode by using the action "Preview" resp. "Edit" in the three-dots-menu in the upper-right corner.

Screenshot

Fixes #352.

### DIFF
--- a/lib/Service/SettingsService.php
+++ b/lib/Service/SettingsService.php
@@ -27,16 +27,7 @@ class SettingsService {
 		$this->l10n = $l10n;
 		$this->root = $root;
 		$this->attrs = [
-			'fileSuffix' => [
-				'default' => '.txt',
-				'validate' => function ($value) {
-					if (in_array($value, [ '.txt', '.md' ])) {
-						return $value;
-					} else {
-						return '.txt';
-					}
-				},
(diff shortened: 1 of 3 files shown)


stefan-niedermann:
Nice :) Honest question: Should this be synvhronized by clients? Keeping the state in sync seems to be a valid goal, but the requirements might be different on a mobile device... 🤔

korelstar:
I thought about that, too. This is why I used exactly the same string labels for the setting and setting options like in your Android app.

With this PR, the new setting will be provided through the (settings) API. Hence, clients can use this setting for their GUI, too. But it's up to the client developer, how this setting should exactly be used in the client app.

Issue #74, opened 2017-03-29
support for attachments / images [$106]


violoncelloCH:
It would be nice to have some kind of attachment support to add an image or another file to a note

---
BountySource is no longer directly supported or endorsed by this repository. See #547 for more information. There is, however, an existing [$136 open bounty](https://www.bountysource.com/issues/43586252-support-for-attachments-images) on this issue.

stefan-niedermann:
If the markdown renderer would support `[image](url)`-format, one could upload an image via drag and drop. The file gets a UUID as name and will be stored in the same folder where the corresponding note file is. After a successfull upload, the notes app can create the markdown with `[original-filename-without-extension](generated-uuid.extension)` and the image gets rendered.

eemantsal:
Any news? Nowadays we take notes with much more content than just text, especially since we all have mobile devices with cameras. I think is a rather basic feature.
There's a desktop application for creating MD notes, Qownnotes, that creates a file named "Media" inside the folder of the note and saves there all the pictures we embed; then, in the note, there's just a link to the corresponding file in that "Media" folder. Seems rather uncomplicated (from a non programmer point of view, heheh).

eemantsal:
The other day I was reading the issues of this application and saw [here](https://github.com/nextcloud/notes/issues/90#issuecomment-299860714) that @stefan-niedermann said that the server app was lacking an API that would add media support for the mobile app.
I thought Hr. Niedermann was referring to the Nextxloud server app, so [there I went to bother NC developers](https://github.com/nextcloud/server/issues/6768) to ask them for media support. @nickvergessen kindly clarified my confusion telling me that is was an NC Notes app issue, not an NC server's, so, here I'm back.

Since I see that @korelstar has already taken this issue into account I'd like to comment an idea from @jancborchardt that he's taken from the IOS app iA Writer, an idea to be considered once NC Notes web app will support media content, of course.
The normal procedure to insert, for example, an image in a Markdown note would be using `![Alt text](image_url)`, but as Mr. Borchardt comments about iA Writer, said app saves all the attached files in the same folder than the .md files, and you just have to type a sort of path to the attached filename preceded, something like `/image.png` (no dot before the slash, just the slash) and here you go.
[...]

PR #785, opened 2021-11-28
Add Imageproxy to resolve relative Imagepaths

newhinton:
Todo, for now see #74 

- [x] Solve upward relative filepaths
- [x] check that only the users own files are accessible
- [ ] ~~allow arbitrary files to be linked, not only images~~ (This is not really supported by markdown afaik)
- [x] fix generation of clientside url for api access
- [x] add upload-api
- [x] add upload-ui
- [ ] ~~maybe add error-image to display anything at all when image not resolvable~~ (not the job of the service)
- [x] fix aaaaaall the stlye issues
- [x] Add Documentation
- [x] imageproxy can't resolve subdirectories (probably because the route does not match anymore with additional /)

~~We should have a serious discussion about the proxy. Currently it allows to get ALL files from files, even outside of the Notes-folder. We may want to restrict access to Files somewhere accessible to the notes app, not all over the instance.~~
Currently only files in the Notes-basefolder are accessible. This removes all security issues posed by allowing abitrary paths.

Also, because of how urls are encoded, `../` is beeing removed (from the url). Therefore, the api only accepts `;;/` as a substitute, and all clients need to replace the  `../` syntax before rendering the link (i am open for suggestions on how to fix this). The markdown-sourcefile **does not need** to be modified, and is not beeing modified by this.

### DIFF
--- a/lib/Controller/NotesController.php
+++ b/lib/Controller/NotesController.php
@@ -341,10 +341,9 @@ class NotesController extends Controller {
 	}
 
 	/**
-	* @NoAdminRequired
-	*/
-	public function createImage($id): JSONResponse
-	{
+	 * @NoAdminRequired
+	 */
+	public function createImage($id): JSONResponse {
 		$file = $this->request->getUploadedFile('image');
 		$result = $this->notesService->createImage(
 			$this->helper->getUID(),
@@ -354,6 +353,4 @@ class NotesController extends Controller {
 
 		return new JSONResponse($result);
 	}
-
-
 }
--- a/lib/Controller/NotesController.php
+++ b/lib/Controller/NotesController.php
@@ -9,6 +9,8 @@ use OCA\Notes\Service\MetaService;
 use OCA\Notes\Service\SettingsService;
 
 use OCP\AppFramework\Controller;
+use OCP\AppFramework\Http\FileDisplayResponse;
+use OCP\Files\IRootFolder;
 use OCP\IRequest;
 use OCP\IConfig;
 use OCP\IL10N;
@@ -29,6 +31,8 @@ class NotesController extends Controller {
 	private $settings;
 	/** @var IL10N */
 	private $l10n;
+	/** @var IRootFolder */
+	private $root;
 
 	public function __construct(
 		string $AppName,
@@ -38,7 +42,8 @@ class NotesController extends Controller {
 		SettingsService $settingsService,
 		Helper $helper,
 		IConfig $settings,
-		IL10N $l10n
+		IL10N $l10n,
+		IRootFolder $root
 	) {
 		parent::__construct($AppName, $request);
 		$this->notesService = $notesService;
@@ -47,6 +52,7 @@ class NotesController extends Controller {
 		$this->helper = $helper;
 		$this->settings = $settings;
 		$this->l10n = $l10n;
+		$this->root = $root;
 	}
 
 	/**
@@ -295,5 +301,44 @@ class NotesController extends Controller {
 			$this->notesService->delete($this->helper->getUID(), $id);
 			return [];
 		});
+	}
+
+	/**
+	 * With help from: https://github.com/nextcloud/cookbook
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 * @return JSONResponse|FileDisplayResponse
+	 */
+	public function getImage(int $id, string $path) {
+		try {
+			$note = $this->notesService->get($this->helper->getUID(), $id);
+
+
+			$notesFolderPath = $this->notesService->getNotesFolder($this->helper->getUID())->getPath();
+			$notePath = $notesFolderPath . "/" . $note->getCategory() . "/";
+
+
+			//because of how the internet works, ../ cannot work in the url and get's removed. We use ;;/ as a placeholder.
+			$parentcount = substr_count($path, ';;/');
+			$path = str_replace(";;/", "", $path);
+
+			$imagepath = $notePath . $path;
+			$origin = $this->root->get($imagepath)->getParent();
+			$notenode = $origin;
+			for ($i = 0; $i < $parentcount; $i++) {
+				$notenode = $notenode->getParent();
+			}
+
+			//return new JSONResponse($notenode->getPath()."-".$origin->getPath()."-".$parentcount);
+
+			foreach ($notenode->getDirectoryListing() as $file) {
+				if (str_ends_with($file->getPath(), $path)) {
+					return new FileDisplayResponse($file, Http::STATUS_OK, ['Content-Type' => 'image/jpeg', 'Cache-Control' => 'public, max-age=604800']);
+				}
+			}
+
+		} catch (\Exception $e) {
+			return new JSONResponse($e->getMessage());
+		}
 	}
 }
--- a/lib/Controller/NotesController.php
+++ b/lib/Controller/NotesController.php
@@ -342,10 +342,12 @@ class NotesController extends Controller {
 					return new DataResponse($notfoundMessage, Http::STATUS_NOT_FOUND);
 				}
 			} catch (Exception $ex) {
-				//this fails rather silently. The exception is not useful aswell. It only occurs when a file does not exist.
+				// this fails rather silently. The exception is not useful. It only occurs when a file does not exist.
 				return new DataResponse($notfoundMessage, Http::STATUS_NOT_FOUND);
 			}
-			return new FileDisplayResponse($targetimage, Http::STATUS_OK, ['Content-Type' => $targetimage->getMimetype(), 'Cache-Control' => 'public, max-age=604800']);
+			$headers = ['Content-Type' => $targetimage->getMimetype(), 'Cache-Control' => 'public, max-age=604800'];
+
+			return new FileDisplayResponse($targetimage, Http::STATUS_OK, $headers);
 		} catch (\Exception $e) {
 			return new DataResponse($notfoundMessage, Http::STATUS_NOT_FOUND);
 		}
--- a/lib/Controller/NotesController.php
+++ b/lib/Controller/NotesController.php
@@ -311,32 +311,35 @@ class NotesController extends Controller {
 	 */
 	public function getImage(int $id, string $path) {
 		try {
+			//decode
+			$path = base64_decode($path);
 			$note = $this->notesService->get($this->helper->getUID(), $id);
 
-
 			$notesFolderPath = $this->notesService->getNotesFolder($this->helper->getUID())->getPath();
-			$notePath = $notesFolderPath . "/" . $note->getCategory() . "/";
+			$notePath = $notesFolderPath . "/";
+			if ($note->getCategory() != "") {
+				$notePath .= $note->getCategory() . "/";
+			}
+
+			//check if targetimage contains subdirectory
+			if(str_contains("/", $path)){
+
+			}
 
 
 			//because of how the internet works, ../ cannot work in the url and get's removed. We use ;;/ as a placeholder.
-			$parentcount = substr_count($path, ';;/');
-			$path = str_replace(";;/", "", $path);
+			$parentcount = substr_count($path, '../');
+			$path = str_replace("../", "", $path);
 
 			$relativeImageNode = $this->root->get($notePath);
 			for ($i = 0; $i < $parentcount; $i++) {
 				$relativeImageNode = $relativeImageNode->getParent();
 			}
+			$targetimage = $this->root->get($relativeImageNode->getPath()."/".$path);
+			return new FileDisplayResponse($targetimage, Http::STATUS_OK, ['Content-Type' => 'image/jpeg', 'Cache-Control' => 'public, max-age=604800']);
 
-			foreach ($relativeImageNode->getDirectoryListing() as $file) {
-				if (str_ends_with($file->getPath(), $path)) {
-					//make sure the file is in the notes folder
-					if (str_starts_with($file->getPath(), $notesFolderPath)) {
-						return new FileDisplayResponse($file, Http::STATUS_OK, ['Content-Type' => 'image/jpeg', 'Cache-Control' => 'public, max-age=604800']);
-					}
-				}
-			}
 		} catch (\Exception $e) {
-			return new JSONResponse($e->getMessage());
+			return new JSONResponse(["This image was not found.", $e->getMessage()]);
 		}
 	}
 
--- a/lib/Controller/NotesController.php
+++ b/lib/Controller/NotesController.php
@@ -333,10 +333,12 @@ class NotesController extends Controller {
 
 			foreach ($notenode->getDirectoryListing() as $file) {
 				if (str_ends_with($file->getPath(), $path)) {
-					return new FileDisplayResponse($file, Http::STATUS_OK, ['Content-Type' => 'image/jpeg', 'Cache-Control' => 'public, max-age=604800']);
+					//make sure the file is in the notes folder
+					if (str_starts_with($file->getPath(), $notesFolderPath)) {
+						return new FileDisplayResponse($file, Http::STATUS_OK, ['Content-Type' => 'image/jpeg', 'Cache-Control' => 'public, max-age=604800']);
+					}
 				}
 			}
-
 		} catch (\Exception $e) {
 			return new JSONResponse($e->getMessage());
 		}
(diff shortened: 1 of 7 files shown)


newhinton:
I added a prototype toolbar for image-uploading. It is barebones, and needs to be improved upon before release, do we want to add additional buttons, similar to how nextcloud/text has one?

in that case i would just hide the toolbar in this pull request, and start working on a proper toolbar after this PR with proper styling and icons.

stefan-niedermann (changes requested):
Could you please add an API documentation to https://github.com/nextcloud/notes/blob/master/docs/api/v1.md ? 🙂

newhinton:
@stefan-niedermann Yes, but i would like to wait for a review of those api-endpoints. I am really not sure if i implemented them in a good way, especially about the `../`-conversion

korelstar:
Nice, that's very cool! Thank you very much for digging into this! I'm very sorry, that I wasn't able to review this, yet. I will do my very best in order to have a deep look on the weekend!

newhinton:
@korelstar No problem! 

It's a bit offtopic, but what do you think about this toolbar idea? The styling is not good at the moment, and some icons are missing, but i think the general idea is not to bad

grafik

newhinton:
@korelstar Could you give me a hint on how i rebuild the css outside of vue files? I'd lile to add icons but they don't seem to update

newhinton:
I have problems with relative subdirectories and the ../ notation.
This is due to how url's are handled by browsers. We also cannot switch to post-requests, since this does not work with the preview. We have to use a get request which encodes the filepath somehow. 

My current idea is to try using base64 encoding, that should get rid of the character issues while easily usable on almost every system that may want to access the api.

newhinton:
base64 encoding works well.
I have also added a local filepicker, however this is more a work in progress, since i have to find out how i can get the app-storage path and the categories to the markdown editor. But it is nice to see it in action!

Also we need to find out if and how we allow both of the picker, local filesystem and the nextcloud one

korelstar:
Thank you very much for your efforts, @newhinton ! I like your approach and I think it's on a good way. It looks that there are some fundamental question to answer resp. decisions to be made. Let's try to collect them.

### How to identify an attachment
You've chosen the file name for identifying the attachment. That's a good, straight forward approach. However, I see one problem: if you add an attachment to your note and then change the note's category, the attachment won't be found anymore. Since I think that's a common use case, we have to find a solution for this.
One approach could be to use the attachment's file id instead of the path. However, this is not very intuitive and could cause problems when migrating notes to another Nextcloud instance.
Another approach would combine both ideas, i.e. file id and path are concatenated, e.g. `629|mypicture.jpg`. But this wouldn't be intuitive, either. The advantage, however, would be that the server could first try to find a file by the file id, and (if this is not successful) then by it's name (or the other way around). If there was a change, the note should be automatically updated (is this feasible?).
Maybe we find another, better approach?

### Source of attachments
Should we allow to attach files to a note from the notes path only or should we allow files from the whole Nextcloud's user path? You've switched to the notes path only for security reasons. I think this is a wise decision, but I want to make this decision clear.

### API path
What about the following in order to eliminate the `../` problem (and render base64 unnecessary):
`GET /notes/{noteid}/attachment?path={path}`
Uploading an attachment could be done with `POST /notes/{noteid}/attachment`.
If we switch to the file id resp. combined approach (see section "How to identify an attachment"), the path would be: `GET /notes/{noteid}/attachment/{attachmentid}?path={path}`

### Toolbar
A toolbar is a good enhancement, let's discuss details in your new issue #788.

### Static CSS
> Could you give me a hint on how i rebuild the css outside of vue files? I'd lile to add icons but they don't seem to update

If you change the scss files in the `css` directory, there is no need for re-building. Building is done by the Nextcloud server automatically.

newhinton:
Could you clarify what you mean with attachment id?

My assumption was that we can only deal with standard "URL's". Those can be relative or absolute, where absolute links are already handled properly.

So we have relative path's which usually point to a local filesystem. So i wanted to resolve this without needing a "new" way to integrate images, as to keep compatibility with other notes apps.

Where would we get an attachement id, and how would we store it?

to your point with `?path=`: if it works, it could be a good idea too. However, quickly rewriting the route results in basically nothing, it redirects to the dashboard. I am not sure if it is supported, because it is not documented: https://docs.nextcloud.com/server/latest/developer_manual/basics/routing.html

newhinton:
I will also remove the toolbar in my cleanup before final commit. It is basically necessary as long as we test/change this code, but the ability to upload files (from the ui) should be included with the toolbar, not this PR

newhinton:
Oh and regarding the issue with switching the category: do other apps take care of this? Should we?

One could argue that it is not our job to take care of that, since other notes could also rely on the image beeing in that folder. If we move it, we could break other notes aswell.

The same argument can be made for deleting images from the source folder.

korelstar:
Attachment id would be the Nextcloud file id of that attachment. But I see your point, such a syntax would not be compatible to other apps. Honestly, I don't know if other apps handle the movement scenario. Maybe, they don't, too. Just wanted to list all issues/decisions that we make, to be clear that we are aware of them.

Regarding the `?path=` thing: You're correct, we can't declare this in the `routes.php`. However, it still works, if `routes.php` declares the url `/note/{id}/attachment` and the respective controller function has a parameter `$path`.

newhinton:
Oh yeah, i tried it without declaring it and it works flawless. 

...i also found the documentation for it: 😀 
https://docs.nextcloud.com/server/latest/developer_manual/basics/controllers.html?highlight=parameter#getting-request-parameters

i will change it to use parameters instead of base64, it is way more intuitive and readable.
I will also assume that `GET /notes/{noteid}/attachment?path={path}` is now "final" and will add it to documentation.

However, one thing: we only render images currently, not other content, so "attachment" is not 100% accurate **at the moment**. We may implement a "download" button for filetypes that we do not recognize in the renderer at a later point.

Though, i like the idea!

korelstar:
> However, one thing: we only render images currently, not other content, so "attachment" is not 100% accurate **at the moment**. We may implement a "download" button for filetypes that we do not recognize in the renderer at a later point.
> 
> Though, i like the idea!

Yes, this should be compatible for future enhancements.

newhinton:
I think i am now ready for a final review. I have refactored a lot and changed naming quite a bit, so it should be way better now. (also the toolbar is gone, but the actual upload/selection functions are still there, they can be easily integrated later)

newhinton:
Oh i remembered something: 

is this the right way to query the current category?
https://github.com/nextcloud/notes/blob/9f0266ccd598a646999245507152829c4f281671/src/components/EditorEasyMDE.vue#L136

i found that it does not work reliable, so this may still be broken in my pr.

Also, is there a discussion on why we call it category and not folder?

newhinton:
grafik

I made a small modification to also render normal files. It downloads the file on click. I do not know if this is in line with how markdown should handle files, but it can easily be removed

korelstar:
> Oh i remembered something:
> 
> is this the right way to query the current category?
> 
> https://github.com/nextcloud/notes/blob/9f0266ccd598a646999245507152829c4f281671/src/components/EditorEasyMDE.vue#L136
> 
> i found that it does not work reliable, so this may still be broken in my pr.

This should be fine. What do you mean exactly with "does not work reliable"? What happens?


> Also, is there a discussion on why we call it category and not folder?

It's our application abstraction layer. For the user, a note is assigned to a category. On the technical backend, a note is stored as a file and the category is stored as folder. But those are different layers. The user does not need to know this. The user is only interested in notes and categories - not in files and folders.

korelstar:
> Oh i remembered something:
> 
> is this the right way to query the current category?
> 
> https://github.com/nextcloud/notes/blob/9f0266ccd598a646999245507152829c4f281671/src/components/EditorEasyMDE.vue#L136
> 
> i found that it does not work reliable, so this may still be broken in my pr.

This should be fine. What do you mean exactly with "does not work reliable"? What happens?


> Also, is there a discussion on why we call it category and not folder?

It's our application abstraction layer. For the user, a note is assigned to a category. On the technical backend, a note is stored as a file and the category is stored as folder. But those are different layers. The user does not need to know this. The user is only interested in notes and categories - not in files and folders.

> I made a small modification to also render normal files. It downloads the file on click. I do not know if this is in line with how markdown should handle files, but it can easily be removed

I like this!

> This branch has conflicts that must be resolved

Could you please fix that conflict? I will do a deeper review later this day.

src/components/EditorMarkdownIt.vue
```diff
@@ -30,13 +37,46 @@ export default {
...
+				}
+
+				token.attrs[aIndex][1] = path
+				// pass token to default renderer.
+				if (path.endsWith("jpg") ||
+					path.endsWith("jpeg") ||
+					path.endsWith("bmp") ||
+					path.endsWith("webp") ||
+					path.endsWith("gif") ||
+					path.endsWith("png")) {
```
korelstar, line 72: The check should ignore the letter case, e.g. it should also work with `JPG` instead of `jpg`.

lib/Controller/NotesController.php
```diff
@@ -296,4 +303,47 @@ public function destroy(int $id) : JSONResponse {
...
+
+			// calculate how many parentnodes we need to get 'up'
+			$parentcount = substr_count($path, '../');
+			$path = str_replace("../", "", $path);
+
+			$relativeImageNode = $this->root->get($notePath);
+			for ($i = 0; $i < $parentcount; $i++) {
+				$relativeImageNode = $relativeImageNode->getParent();
+			}
+			$targetimage = $this->root->get($relativeImageNode->getPath()."/".$path);
```
korelstar, line 330: There must be a check, if we are still in the notes path. Currently, I can load images from other users!

korelstar (changes requested):
I finally found some time for a deep review :-)
First of all: thank you again for your work on this! I'm writing some general remarks and I did commented some special changes.

### Code style
Please use `make lint` for checking your code style and fix all errors and warnings regarding your changes.

### external API
Currently, you've implemented the internal API (`NotesController`). That's fine and sufficient for now. However, you've documented the external API (`docs/api/v1.md`) although it is not implemented yet. I think we can postpone this for the next step. Since you're developing very fast and you have opened multiple PRs that are partly conflicting, let's concentrate on small parts.

### Controller vs. Service
That controllers are used to connect routes with app logic. But please note, that the actual app logic should be in respective service classes. On the other hand, the service should not contain any code regarding the provided API (e.g. `Http` stuff), this has to be done in the controller.
To be more concrete: please move most of the method `NotesController.getAttachment` (except the `Response` stuff) to `NotesService` and move the `Response` stuff from  `NotesService.createImage` to `NotesController`.

lib/Controller/NotesController.php
newhinton: ohh yeah, just to understand how you accessed this:

did you prepend '../' the exact same amount to logically get back to "root"? and then append a full, internal nextcloud-path including the different user?

korelstar: Yes

src/components/EditorMarkdownIt.vue
```diff
@@ -14,14 +15,20 @@ export default {
...
+		const markdown = new MarkdownIt({
+			linkify: true,
+		})
+
 		return {
 			html: '',
-			md: new MarkdownIt({
-				linkify: true,
-			}),
+			md: markdown
```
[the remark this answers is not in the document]
newhinton: Sorry, this was leftover from my initial trials for rewriting the url

lib/Controller/NotesController.php
newhinton: Could you provide me with your testsetup? 

I know why and how to fix it, but i cannot replicate it since i always get a second / in my path
Eg.: `?path=../../../test/files/Nextcloud Manual.pdf` results in an exception to be thrown, not an image to be served.

newhinton: I think i fixed it, but since i could not properly reproduce it i cant really verify it. Please try to break it again! I want to be sure this cannot be abused

korelstar: My test setup:
User `A` has a note in `notes/test/test.md` with the following content:
```
![Attack](../../../../B/files/DSC_0020.jpg)
```
User `B` has the image file `DSC_0020.jpg` in his user directory.

I will test your changes soon.

src/components/EditorMarkdownIt.vue
```diff
@@ -61,6 +61,7 @@ export default {
 				}
 
 				token.attrs[aIndex][1] = path
+				path = path.toLowerCase()
```
korelstar, line 64: This breaks the download if the file name has upper case letters. Please use a new variable (e.g. `lowecasePath`) for this.

lib/Controller/NotesController.php
korelstar: Thanks, looks like your fix is working. I think your solution on how to fix this is a good approach! :+1:

korelstar:
@newhinton What's the state of this PR? There is still a merge conflict. Could you please rebase this branch in order to solve the conflict?

korelstar:
You will always have to force push after rebasing. But I'm also fine if you create a new branch. Thanks for your efforts!

newhinton:
okay then i'll force push! Will probably do so tomorrow. Also: We should probaby check the service very carefully again, just in case. If i got this wrong, there is a huge hole in nextclouds security

newhinton:
Done! This branch is now up to date.

newhinton:
@korelstar did you mean the DCO?

newhinton:
Interesting that they did not show up in the testruns. Anyway, i fixed them except those i did not cause, and a single one i dont know how to fix:

The file we are serving is actually not a `File`, but a `Node`. The `FileDisplayResponse` expects files, so it complains. It works because `File` extends `Node`. Just casting did not seem to work. How do i get a file from a node?

korelstar:
This can be done by inserting `assert($targetimage instanceof \OCP\Files\File);`

I added this and I also did some of the major remarks from https://github.com/nextcloud/notes/pull/785#pullrequestreview-829537246 . Will continue with reviewing later.

newhinton:
@korelstar Actually i do have a question regarding my changes to the readme. Since the imageproxy is available through an endpoint, doesn't that count as an external api? Why should it not be documented?

korelstar:
> @korelstar Actually i do have a question regarding my changes to the readme. Since the imageproxy is available through an endpoint, doesn't that count as an external api? Why should it not be documented?

The notes app has two REST interfaces: one for the web interface (`NotesController`, internal API) and one for third-party apps (`NotesApiController`, external API). The documentation is about the external API.

In your current implementation, the proxy is not available from external, since it uses the `NotesController`. Therefore we must not document this in the external API documentation.

Of course, the goal is to provide this new feature also to the external API. But let's do this in a separate PR, so we can merge this PR earlier and do further tests with it. Another point: when changing the external API, you will have to write respective tests. This is currently missing.

korelstar (approved):

korelstar:
Just did some last improvements. If you agree, I will merge this now.

newhinton:
The simplification looks good to me, except i cant see how we make sure we did not leave the notes directory. Was the removal of the last if clause wether the resulting node is still in the notes-path intentional?

korelstar:
> Edit: I think i figured it out, you do not actually work on a full path, but only on the categories of a note. So when we "overstep" the boundaries we get 'nothing' (in terms of the array you are using) and then we get the child of the notes folder with that array

Exactly!

korelstar:
Next steps are:

## 1. Add upload feature to web-frontend
Implement using toolbar (requires #790) or at the first step with a new entry in the three-dot menu. I vote for the last since we don't have any dependencies, then. Would you like to work on this?

## 2. Public API for third-party apps
- discuss public API with third-party app developers (what are their desires?)
- add new API methods to `NotesApiController`
- add API documentation (here is the last state: https://github.com/nextcloud/notes/blob/aac2223717065fa216f7be315df3bcde332d7c3a/docs/api/v1.md)
- add API tests (see here: https://github.com/nextcloud/notes/blob/master/tests/api/APIv1Test.php)

newhinton:
@korelstar would you mark #74 as resolved? I'd like to claim the bounty for the ticket

newhinton:
oh and i will continue work on the toolbar aswell, though that might take a couple of weeks

korelstar:
> @korelstar would you mark #74 as resolved? I'd like to claim the bounty for the ticket but i dont think its my place to close the ticket on my own for that

I don't think that it's already solved. But you're on a good way! A big part of that issue is about uploading a new image. You've already prepared this, but it's currently not usable. The original issue's requirements for uploading are here: https://github.com/nextcloud/notes/issues/74#issuecomment-290219121 . In addition, you could add an upload menu item (see https://github.com/nextcloud/notes/pull/785#issuecomment-1025143952).

Issue #722, opened 2021-05-25
Ability to change default file extension


pprevos:
I am n Emacs Org-Mode user and would love to be able to select ".org" as an extension.

It would be great to have the ability to change or add default extensions (currently only .md or .txt) through a config file.

The alternative I am working with is to convert the .md notes to Org in Emacs. I can view Org files made in Emacs in NextCloud, but cannot create new .org files with NextCloud.

Org-Mode is growing in popularity and non-Emacs-systems, such as Hugo and GitHub are now supporting this format.

No biggie, just a nice to have.

stefan-niedermann:
@korelstar please note, that this is explicitly **not** a request to support `org-mode` as editor or viewer ([i already talked to the author](https://help.nextcloud.com/t/writing-notes-in-org-mode/116993/2?u=stefan-niedermann)). It is just a request to make the available file extension options configurable via a config file (where `.txt` and `.md` should stay the default options of course).

korelstar:
I see. Since we already accept files with this extension as notes, I think it's okay to allow this to be set as default extension.

Should we then allow to set each of the accepted extensions? See https://github.com/nextcloud/notes/blob/v4.0.4/lib/Service/NotesService.php#L174

korelstar:
But should we really differentiate between config file and GUI? How could users find the config file option? Shouldn't we instead provide all allowed extensions in the drop-down?

On the other hand, when implementing this using the config file approach, this would fix https://github.com/nextcloud/notes/issues/629, too.

Any opinions? I have to think about it.

stefan-niedermann:
Huh, i wasn't even aware that all of those extensions are already recognized :smile:... Good question.
I think it will be simpler for users to choose between `.txt` and `.md` plus explicitly custom extensions which have been configured explicitly by the user as requested in this issue.
I wouldn't offer all of those as available extensions by default though, even if the app is able to recognize them.

I understand from a maintenance point of view that you don't want to have multiple lists, but from a UX point of view, it's simply overwhelming for a user, especially because `markdown` and `note` are pretty uncommon file extensions...

Will also think about it... and i guess i need a REST endpoint to fetch the available extensions if the list isn't static anymore...

korelstar:
I still have a problem with restricting the user choice to `.txt` and `.md` and -- at the same time -- allow the user to chose other extensions using a more complicated way. In fact, the "more complicated way" means, that a normal user can't use it, but you need to be an admin in order to edit the config file.

[...]

PR #795, opened 2021-12-16
Add .org to notes extensions list

vincowl:
Relates to #722

### DIFF
--- a/lib/Service/NotesService.php
+++ b/lib/Service/NotesService.php
@@ -171,7 +171,7 @@ class NotesService {
 	/**
 	 * test if file is a note
 	 */
-    public function isNote(FileInfo $file, string $userId) : bool {
+	public function isNote(FileInfo $file, string $userId) : bool {
 		static $allowedExtensions = ['txt', 'org', 'markdown', 'md', 'note'];
 		$customExtension = $this->getCustomExtension($userId);
 		$ext = strtolower(pathinfo($file->getName(), PATHINFO_EXTENSION));
--- a/lib/Service/NotesService.php
+++ b/lib/Service/NotesService.php
@@ -107,7 +107,7 @@ class NotesService {
 		// get file name
 		$fileSuffix = $this->settings->get($userId, 'fileSuffix');
 		if ($fileSuffix === "custom") {
-			$fileSuffix = $this->settings->get($userId, 'suffixValue');
+			$fileSuffix = $this->settings->get($userId, 'customSuffix');
 		}
 		$filename = $this->noteUtil->generateFileName($folder, $title, $fileSuffix, -1);
 		// create file
@@ -187,8 +187,8 @@ class NotesService {
 	 * Update the value of user defined files extension
 	 */
 	private function setCustomExtension(string $userId) {
-		$suffix = $this->settings->get($userId, 'suffixValue');
-		$this->customExtension = strtolower(ltrim($suffix, "."));
+		$suffix = $this->settings->get($userId, 'customSuffix');
+		$this->customExtension = ltrim($suffix, ".");
 	}
 
 	/**
--- a/lib/Service/NotesService.php
+++ b/lib/Service/NotesService.php
@@ -113,9 +113,7 @@ class NotesService {
 		// create file
 		$file = $folder->newFile($filename);
 
-		$note = new Note($file, $notesFolder, $this->noteUtil);
-
-		return $note;
+		return new Note($file, $notesFolder, $this->noteUtil);
 	}
 
 
@@ -180,7 +178,7 @@ class NotesService {
 	 * test if file is a note
 	 */
 	public function isNote(FileInfo $file) : bool {
-		static $allowedExtensions = ["txt", "org", "markdown", "md", "note"];
+		static $allowedExtensions = ['txt', 'org', 'markdown', 'md', 'note'];
 		$ext = strtolower(pathinfo($file->getName(), PATHINFO_EXTENSION));
 		return $file->getType() === 'file' && (in_array($ext, $allowedExtensions) || $ext === $this->customExtension);
 	}
--- a/lib/Service/NotesService.php
+++ b/lib/Service/NotesService.php
@@ -13,8 +13,6 @@ class NotesService {
 	private $settings;
 	private $noteUtil;
 
-	private $customExtension = "";
-
 	public function __construct(
 		MetaService $metaService,
 		SettingsService $settings,
@@ -26,9 +24,8 @@ class NotesService {
 	}
 
 	public function getAll(string $userId) : array {
-		$this->setCustomExtension($userId);
 		$notesFolder = $this->getNotesFolder($userId);
-		$data = $this->gatherNoteFiles($notesFolder);
+		$data = $this->gatherNoteFiles($userId, $notesFolder);
 		$fileIds = array_keys($data['files']);
 		// pre-load tags for all notes (performance improvement)
 		$this->noteUtil->getTagService()->loadTags($fileIds);
@@ -53,9 +50,8 @@ class NotesService {
 	}
 
 	public function get(string $userId, int $id) : Note {
-		$this->setCustomExtension($userId);
 		$notesFolder = $this->getNotesFolder($userId);
-		$note = new Note($this->getFileById($notesFolder, $id), $notesFolder, $this->noteUtil);
+		$note = new Note($this->getFileById($userId, $notesFolder, $id), $notesFolder, $this->noteUtil);
 		$this->metaService->update($userId, $note);
 		return $note;
 	}
@@ -99,7 +95,6 @@ class NotesService {
 	 */
 	public function create(string $userId, string $title, string $category) : Note {
 		// get folder based on category
-		$this->setCustomExtension($userId);
 		$notesFolder = $this->getNotesFolder($userId);
 		$folder = $this->noteUtil->getCategoryFolder($notesFolder, $category);
 		$this->noteUtil->ensureSufficientStorage($folder, 1);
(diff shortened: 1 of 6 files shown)


korelstar (changes requested):
Thanks for your contribution! I thought about this approach and had a look (again) at the org-mode "standard". Finally, I came to the conclusion, that adding `org` as option for the default extension doesn't make any sense for me. The org-mode markup is completely different to Markdown. I even thought about removing `org` from the list of allowed extensions. However, I will keep that list as it is for backward-compatibility reasons. Nevertheless, we will not provide another markup style in this app. Hence, promoting `org` as default extension is not an option for me.

Instead, we could allow any user-defined extension. What about providing three file suffix options: `.txt`, `.md` and `user defined`. When choosing `user defined`, another input field could be shown (with the current file suffix as default value) which allows users to define arbitrary suffixes. Note: we would need to dynamically add this user-defined suffix to the list of allowed extensions in `NotesService.isNote()`.

What about that approach? /cc @nextcloud/notes
Would you be willing to implement that? @vincowl

vincowl:
> Thanks for your contribution! I thought about this approach and had a look (again) at the org-mode "standard". Finally, I came to the conclusion, that adding `org` as option for the default extension doesn't make any sense for me. The org-mode markup is completely different to Markdown. I even thought about removing `org` from the list of allowed extensions. However, I will keep that list as it is for backward-compatibility reasons. Nevertheless, we will not provide another markup style in this app. Hence, promoting `org` as default extension is not an option for me.

That's fine to me. It perfectly makes sense and there is no reason to support org-mode markup in notes

> Instead, we could allow any user-defined extension. What about providing three file suffix options: `.txt`, `.md` and `user defined`. When choosing `user defined`, another input field could be shown (with the current file suffix as default value) which allows users to define arbitrary suffixes. Note: we would need to dynamically add this user-defined suffix to the list of allowed extensions in `NotesService.isNote()`.
> What about that approach? /cc @nextcloud/notes Would you be willing to implement that? @vincowl

Looks great. I don't have much time but I can try to implement this. The use case is clear. Just to be sure : when you say `which allows users to define arbitrary suffixes`, we can only add one suffix ? The `es` ending suffix can be error-prone

korelstar:
> Just to be sure : when you say `which allows users to define arbitrary suffixes`, you mean 'the user can only add one suffix' ? The `es` ending suffix can be error-prone

Yes, one should define one custom suffix only.

vincowl:
I just started today to code this PR. Everything seems to work, except the part related to the dynamical addition of the user-defined suffix to the list of allowed extensions in `NotesService.isNote()`. 

Here is what I did : 

- define a static class property in NotesService : 
```php 
private static $customExtension;
```
- Update the code of `isNote()` to 
```php
private static function isNote(FileInfo $file) : bool {
    static $allowedExtensions = ["txt", "org", "markdown", "md", "note"];
    $ext = strtolower(pathinfo($file->getName(), PATHINFO_EXTENSION));
    return $file->getType() === 'file' && (in_array($ext,$allowedExtensions) || $ext === self::$customExtension);
}
```
- create a methode in `NotesService` : 
```php
/**
* Update the value of user defined files extension
*/
public static function setCustomExtension(string $suffix) {
    self::$customExtension = ltrim($suffix, ".");
}
```
- Call this method when the user updates the extension. This avoids to pop / push extensions to the allowed list that I cannot access outside of `isNote`

The problem is that the `$customExtension` isn't updated in `isNote` as if the static method wouldn't check the property's new value and would keep only it's initial one.
Any idea ? @korelstar 

Second trouble would be 'how to intialize the static property with stored settings value on 
 a new session' ?

vincowl:
Committing the current state of code if needed

korelstar:
Hey @vincowl , thanks for working on this! Looks like a good start! Please keep the Nextcloud coding standard mind. We use 1 tab for indentation. You can check your code by executing `make lint` (or `make lint-php` if you are looking on PHP only).

Regarding your question: please note that in PHP, static variables belong to the current HTTP request only. Therefore, you will have to fill the variable on every request. I suggest to do remove the `static` attribute from `$customExtension` and from `isNote()`. Then, you can fill `$customExtension` in the `NotesService`'s contructor.

One more thing: I'm not sure if we need to create a new settings attribute `suffixValue`. It would be more elegant, if the existing attribute `fileSuffix` saves the actual (custom) file suffix and not just the value "custom". Of course, the attribute definition (`SettingsService->attrs`) has to be changed.

vincowl:
@korelstar I'll check the coding standards before submitting the PR.

I am not sure to fully understand the lifecycle of the `NotesService` objects : are they created at each HTTP request ? If so, does the `settings` property contain the current saved settings when creating the `NotesService` object ?

If I use the `fileSuffix` attribute to store the custom suffix, then I won't be able to retrieve my custom suffix in the following use case :
- I need to create notes in .org for instance, so I choose `custom` and set `fileSuffix` to `.org`
- I create notes
- I need to work now with .md notes, so I change the `fileSuffix` to `.md`
- then I need to be back to .org : I need to retype my extension. If I had stored my custom extension in a new attribute, then it would be filled up automatically, whatever the lifecycle would have been.

Shall I still remove the `suffixValue` ? Since this PR is just a 'convenience feature', I coded this to have the less impact on the 'core features' as possible, so I chose to avoid modifying actual attributes.

vincowl:
@korelstar This should be ok. I have a working solution here.

I didn't manage to set the `customExtension` value in the constructor since it would have impacted the whole calls to the constructor since I need the `userId` parameter to retrieve stored settings.

korelstar:
Sorry for the late reply!

> I didn't manage to set the `customExtension` value in the constructor since it would have impacted the whole calls to the constructor since I need the `userId` parameter to retrieve stored settings.

But then it's very error prone to make `$customExtension` an object variable, because there may be cases when `NotesService->customExtension`  is not set but the caller forgot to call `setCustomExtension()`. Instead, you should introduce `$customExtension` as a new method parameter for `isNote` (and consequently also for `getFileById` and `gatherNoteFiles`; which then can stay `static` btw.). But the caller methods (`getAll`, `get` and `delete`) must call a new method `getCustomExtension(string $userId)` for filling the new parameter.

By the way: you are doing a `strtolower()` with the custom extension multiple times. I think we should remove this completely, since upper case may be intended.

> Shall I still remove the `suffixValue` ? Since this PR is just a 'convenience feature', I coded this to have the less impact on the 'core features' as possible, so I chose to avoid modifying actual attributes.

I still vote for removing `suffixValue`, because I think it will be more error prone, if the real suffix is sometimes in `fileSuffix` and sometimes in `suffixValue`. IMHO that's not a good design.

vincowl:
Fine. Do you realize how these changes will modify your interfaces, will impact frontend logics and will engrave a side feature that was considered as an unwanted one in your whole design ? Maybe it's not the cleanest design but it minimizes the impact on the "core code". 

I don't know vue.js and will try to do my best to implement this even if I lack docs or comments. 

If you mix suffixes in the same variable in the frontend, it also implies we will systematically add the suffix to the extensions list in isNote. Right ?

korelstar:
Maybe you're right, and we should do the easy way first. We could refactor that later, when custom suffixes has been established. So you can skip removing `suffixValue`, but please rename it to `customSuffix` (or something similar) - then it's clearer which of the both attributes has which semantic.

lib/Service/NotesService.php
```diff
@@ -169,19 +171,28 @@ private static function gatherNoteFiles(Folder $folder, string $categoryPrefix =
 	/**
 	 * test if file is a note
 	 */
-	private static function isNote(FileInfo $file) : bool {
+    public function isNote(FileInfo $file, string $userId) : bool {
```
korelstar: Indentation is not correct (see lint) and public is not needed:
```suggestion
	private function isNote(FileInfo $file, string $userId) : bool {
```

lib/Service/SettingsService.php
```diff
@@ -47,6 +47,16 @@ public function __construct(
...
 			'noteMode' => $this->getListAttrs('edit', 'preview'),
+			'customSuffix' => [
+				'default' => "",
+				'validate' => function ($value) {
+					$out = "";
+					if (preg_match('/^\.[A-Za-z0-9.-]*$/', $value)) {
+						$out = $value;
+					}
+					return $out;
+				},
```
korelstar, line 58: I did some tests with unexpected inputs. In this case, it results in an empty extension, which is inconvenient. What about the following solution?
1. it does not block wrong inputs, but removes illegal characters only.
2. if the result is empty, it uses `.txt` as default.

```suggestion
				'default' => '.txt',
				'validate' => function ($value) {
					$out = ltrim(preg_replace('/[^A-Za-z0-9.-]/', '', $value), '.');
					if (empty($out)) {
						$out = 'txt';
					}
					return '.' . $out;
				},
```

korelstar (changes requested):
Thanks! Looks really good, now!

> but instead of calling `getCustomExtension` in `getFileById` and in `gatherNoteFiles`, I made this call only once in `isNote`, which forbids to declare static the previous methods. I didn't clearly understand the importance of making them static, so I thought this was cleaner to minimize the calls to `getCustomExtension`. If it is better to declare them static, let me know and I'll modify the code.

The `static` property is not important. But with your approach, `getCustomExtension` is called unnecessarily often (imagine a call to `getAll` with 1.000 notes, here `getCustomExtension` is called 1.000 times). This would be optimized with my approach. If you're willing, please change it. But the other remark (input validation) is more important :-)

lib/Service/SettingsService.php
vincowl: Looks fine to me ! I didn't realize that a "wrong" extension would result in an empty one. Nice catch !

lib/Service/NotesService.php
vincowl: my commit correcting linting error and your review collided. Fully agree to make this function private

vincowl:
I modified the calls to `getCustomExtension`. Indeed I didn't trace back the calls to this method in `getAll`... You're totally right, that's a bad design from me.

vincowl:
This PR should resolve #722

korelstar (approved):

PR #821, opened 2022-02-19
fix noteservice not returning the final filename

newhinton:


newhinton:
This is required so that applications can properly create the markdown since the filename is generated serverside

korelstar (approved):

PR #834, opened 2022-03-19
Improvements on images/attachments

korelstar:
Follow-up of #785 and #823

- [x] show images that are included from the Text app
- [x] fix images with space in file name
- [x] improve design


korelstar:
CI tests currently fail due to a bug in Guzzle, see https://github.com/guzzle/guzzle/issues/2998, https://github.com/guzzle/psr7/issues/489 and https://github.com/guzzle/psr7/issues/490

Issue #413, opened 2019-12-01
Docu for Categories / category handling


MarcusE1W:
The request is to document how categories work with notes, maybe in the Wiki.

Currently I was struggling a bit how to **create** categories (using an iPad and Firefox browser).
Searching the Notes app GUI I could not figure out how to create a category in the notes app.

Eventually I tried to just create a new folder in the files app. Still no luck. However after I have moved one note to the new folder the new category also shows up in the notes app. I am not quite sure if this is how categories are supposed to be created ?

Even after a new category is now shown in notes I can’t **move** the files between categories in the notes app, although there seems to be a option in the drop down menu in the sidebar, however only the current category is shown. Should moving categories also be done in the files app?

Once the suggested way to use categories is clarified I am happy to write a Wiki entry.

raelb:
I too was perplexed as to how to create categories, expecting there to be a create button of some sort.

When there are no notes there is some information provided about how to use the app. Perhaps in that screen, some comment of how to create categories would be helpful.

korelstar:
Thanks for the feedback! I think there are two aspects in your feedback:

## 1. Does changing categories work as expected?

The current workflow for setting a category is as follows:

- open the note in the notes app
- open "details" using the three-dot menu in the upper right corner
- you see the details sidebar on the right:
![Screenshot](https://user-images.githubusercontent.com/6277619/70847923-8ed65800-1e6a-11ea-8371-a268292267c8.png)
- when clicking on the current category, you can chose another category for the current note:
![Screenshot](https://user-images.githubusercontent.com/6277619/70847926-a01f6480-1e6a-11ea-8ba4-8c4818796f61.png)
- you can even type a new category name in that field in order to create and set a new category:
![Screenshot](https://user-images.githubusercontent.com/6277619/70847931-b6c5bb80-1e6a-11ea-85c2-e0a9c6cd88c8.png)

If this workflow doesn't work in your installation, please create a new issue for that.

## 2. How does a user get in touch with that workflow?

I agree that we should improve this. Based on your feeback, I propose these changes:

- Add "details" action to the note's three-action menu in the app-navigation (left side). With this, the "details" sidebar can be found easier.
- Add short explanation on the first-use-screen (shown when there are no notes in your profile).

I don't know if this is sufficient. If you have some other ideas, please write them down in this issue.

[...]

PR #839, opened 2022-03-25
Added option to have all categories always visible in the navigation list

joachimeichborn:
I found it very uncomfortable to work with categories as they are always collapsed whenever I switch between them in the navigation list. Therefore I have added an option to the settings to have them always visible. If this is turned off (Settings -> Display mode for categories -> Collapse categories) everything behaves as before. If it is turned on (Settings -> Display mode for categories -> Always show all categories) a list of all categories is always shown with the active category being highlighted.

Implementation wise I have extracted the existing categories list from NavigationCategoriesItem to NavigationCategoriesList. It is then either used inside NavigataionCategoriesItem to show the uncollapsed list as before or directly inside NavigationCategoriesList to show the always visible category list (depending on the chosen setting).

### DIFF
--- a/l10n/de_DE.js
+++ b/l10n/de_DE.js
@@ -20,6 +20,8 @@ OC.L10N.register(
     "File extension for new notes" : "Dateiendung für neue Notizen",
     "Display mode for notes" : "Anzeigemodus für Notizen",
     "User defined" : "Benutzerdefiniert",
+    "Always show all categories" : "Immer alle Kategorien anzeigen",
+    "Collapse categories" : "Kategorien einklappen",
     "Open in edit mode" : "Im Bearbeitungsmodus öffnen",
     "Open in preview mode" : "Im Vorschaumodus öffnen",
     "No notes yet" : "Noch keine Notizen vorhanden",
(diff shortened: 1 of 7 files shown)


stefan-niedermann:
@korelstar wouldn't it make sense to rethink the default and simply *show* the categories? I agree that category usage is currently not optimal and in my opinion we could simply show them always without an option 🤷

korelstar:
I agree, too, that the current category handling is sub-optimal. But there is already a discussion on this, see #413. I apologize, that this is still not fixed (due to lack of time). But the goal is clear: We want to switch to a three-column layout using the standard components from [`nextcloud-vue`](https://github.com/nextcloud/nextcloud-vue):

- [`AppNavigation`](https://nextcloud-vue-components.netlify.app/#/Components/App%20containers/AppNavigation?id=appnavigation-1) contains the (opened) category list (and the settings at the bottom)
- [`AppContentList`](https://nextcloud-vue-components.netlify.app/#/Components/App%20containers/AppContent?id=appcontentlist) contains the list of notes (using [`ListItem`](https://nextcloud-vue-components.netlify.app/#/Components/List%20items?id=listitem) component)
- [`AppContentDetails`](https://nextcloud-vue-components.netlify.app/#/Components/App%20containers/AppContent?id=appcontentdetails) is the current note editor

With this goal in mind, I don't think that it is productive to introduce an interim solution like this PR. @joachimeichborn, I would appreciate very much if you could help with implementing the final solution! Maybe, you would like to extend this PR or start a new PR on this? That would be very cool!

korelstar:
Closing in favor of #842

PR #859, opened 2022-05-21
allow double click to change to editing

### LINKED ISSUE #857: Double Click action in View mode
It could be good if a note would switch to editing mode if a user double clicks the view.

I am willing to implement this, this ticket is more a discussion if that is something that would actually make the ui better.

newhinton:

fixes #857

### DIFF
--- a/src/components/Note.vue
+++ b/src/components/Note.vue
@@ -4,7 +4,7 @@
 			id="note-container"
 			class="note-container"
 			:class="{ fullscreen: fullscreen }"
-			 v-on:dblclick="onDoubleClick"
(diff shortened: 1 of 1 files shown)


korelstar (changes requested):
See also: https://github.com/nextcloud/notes/issues/857#issuecomment-1140444915

korelstar:
Okay, I finally think we should not introduce this, since double clicking text has another meaning in browsers (see https://github.com/stefan-niedermann/nextcloud-notes/issues/1519#issuecomment-1167072504).

PR #861, opened 2022-05-22
Feature/858/checkboxtoggle

### LINKED ISSUE #858: Allow Checkbox to be toggled in viewmode
A clear and concise description of what the problem is. Ex. I'm always frustrated when i cant quickly change a checkbox

Changeable checkboxes in viewmode.

a clear and concise description of any alternative solutions or features you've considered
No changes to the viewmode

**Additional context**
I am willing to implement this together with #857

newhinton:
fixes #858

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,7 +4,6 @@
 	"requires": true,
 	"packages": {
 		"": {
-			"name": "notes",
 			"license": "agpl",
 			"dependencies": {
 				"@nextcloud/axios": "^1.9.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,7 @@
 	"requires": true,
 	"packages": {
 		"": {
+			"name": "notes",
 			"license": "agpl",
 			"dependencies": {
 				"@nextcloud/axios": "^1.9.0",
--- a/src/components/EditorMarkdownIt.vue
+++ b/src/components/EditorMarkdownIt.vue
@@ -64,7 +64,6 @@ export default {
 			}
 			const items = document.getElementsByClassName('task-list-item')
 			for (let i = 0; i < items.length; ++i) {
-				console.log(items[i])
 				items[i].removeEventListener('click', this.setListener)
 				items[i].addEventListener('click', this.setListener)
 			}
@@ -86,7 +85,7 @@ export default {
 		},
 
 		checkLine(line, index, id, target) {
-			let returnValue = line;
+			let returnValue = line
 			if ('cbx_' + id === target.id) {
 				if (target.checked) {
 					returnValue = returnValue.replace('[ ]', '[x]')
--- a/src/components/EditorMarkdownIt.vue
+++ b/src/components/EditorMarkdownIt.vue
@@ -41,7 +41,7 @@ export default {
 
 	watch: {
 		value: 'onUpdate',
-		html: 'onChange'
+		html: 'onChange',
 	},
 
 	created() {
@@ -55,28 +55,28 @@ export default {
 		},
 
 		onChange() {
-			const markdown = this.value;
+			const markdown = this.value
 			const updateFunction = this.updateMarkdown
-			const items = document.getElementsByClassName('task-list-item');
+			const items = document.getElementsByClassName('task-list-item')
 			for (let i = 0; i < items.length; ++i) {
 				items[i].addEventListener('click', function() {
-					const wasChecked = !this.children[0].checked;
+					const wasChecked = !this.children[0].checked
 
-					let idOfCheckbox = 0;
-					const markdownLines = markdown.split( '\n');
+					let idOfCheckbox = 0
+					const markdownLines = markdown.split('\n')
 					markdownLines.forEach((line, i) => {
-						if(line.trim().startsWith("- [x]") || line.trim().startsWith("- [ ]")){
-							if("cbx_"+idOfCheckbox === this.children[0].id){
-								if(wasChecked) {
-									markdownLines[i] = markdownLines[i].replace("[x]","[ ]")
+						if (line.trim().startsWith('- [x]') || line.trim().startsWith('- [ ]')) {
+							if ('cbx_' + idOfCheckbox === this.children[0].id) {
+								if (wasChecked) {
+									markdownLines[i] = markdownLines[i].replace('[x]', '[ ]')
 								} else {
-									markdownLines[i] = markdownLines[i].replace("[ ]","[x]")
+									markdownLines[i] = markdownLines[i].replace('[ ]', '[x]')
 								}
 							}
-							idOfCheckbox++;
+							idOfCheckbox++
 						}
-					});
-					updateFunction(markdownLines.join('\n'));
+					})
+					updateFunction(markdownLines.join('\n'))
 				})
 			}
 		},
--- a/src/components/EditorMarkdownIt.vue
+++ b/src/components/EditorMarkdownIt.vue
@@ -75,7 +75,7 @@ export default {
 			markdownLines.forEach((line, i) => {
 				// Regex Source: https://github.com/linsir/markdown-it-task-checkbox/blob/master/index.js#L121
 				// plus the '- '-string.
-				if (/^(- )\[[xX \u00A0]\][ \u00A0]/.test(line.trim())) {
+				if (/^-\s+\[[xX \u00A0]\][ \u00A0]/.test(line.trim())) {
 					markdownLines[i] = this.checkLine(line, i, idOfCheckbox, item.target)
 					idOfCheckbox++
 				}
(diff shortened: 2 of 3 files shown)


package-lock.json
```diff
@@ -4,7 +4,6 @@
 	"requires": true,
 	"packages": {
 		"": {
-			"name": "notes",
```
korelstar, line 7: Is this intended? If not: please remove this change (this will reduce conflicts)

src/components/EditorMarkdownIt.vue
```diff
@@ -52,6 +53,40 @@ export default {
...
+			const updateFunction = this.updateMarkdown
+			const items = document.getElementsByClassName('task-list-item')
+			for (let i = 0; i < items.length; ++i) {
+				items[i].addEventListener('click', function() {
+					const wasChecked = !this.children[0].checked
+
+					let idOfCheckbox = 0
+					const markdownLines = markdown.split('\n')
+					markdownLines.forEach((line, i) => {
+						if (line.trim().startsWith('- [x]') || line.trim().startsWith('- [ ]')) {
```
korelstar, line 68: Does this respects every possible syntax? E.g. capital `X` vs. small `x`, additional spaces after the dash `-`, non-breaking space instead of normal space. See also https://github.com/linsir/markdown-it-task-checkbox/blob/master/index.js#L121

korelstar (changes requested):
Thanks for this PR! I really like it! Here are some remarks for improving it:

- Changing a checkbox only works once for me. After that, any other change is done visually, but not in the note. I have to reload to proceed.
- Please respect the read-only attribute. The checkboxes should not be editable if read-only mode is active.
- Could you please change the mouse cursor? Currently it is text selection, but I think it should be the default pointer.
- Would it be possible to use the same checkbox design like in edit mode?

package-lock.json
newhinton: Yeah my IDE always wants to add that. This time it slipped, i will remove it.

newhinton:
>Changing a checkbox only works once for me. After that, any other change is done visually, but not in the note. I have to reload to proceed.

That is interesting. I can reproduce, however, the onEdit call does take the correct content, but it seems it does not save it. I have to figure out why.

>Would it be possible to use the same checkbox design like in edit mode?
Yes, but no. :D I have already changed the checkbox design in #790 to use rounded corners, so that the editor matches the preview. If you dont object, i would like to do it that way:

Editor:
grafik

Preview:
grafik
 (the editor has a slightly smaller border)

newhinton:
How do i test the read-only attribute? I have passed it just like the editor so it 'should' be fine, but i have never encountered a read only note. Are those shared from other people?

src/components/EditorMarkdownIt.vue
newhinton: I have switched to the regex aswell to make sure it works

newhinton:
So, i fixed the single click only bug. Sadly i had to introduce a timeout function so that the v-html can render and the new content is actually found with the checkboxes. If you have a better idea please tell me.


Also i found a very strange bug, but i think this is the issue with MarkdownIT.

```
 - [x] test
 - [x] test2
 [ ] test
 
 
 test
```

This note will have 2 checkboxes, as expected, but the label of the second one extends to a second line. Can you confirm this behavioud on your end?

src/components/EditorMarkdownIt.vue
```diff
@@ -51,7 +55,49 @@ export default {
...
+			}
+		},
+
+		setListener(item) {
+			let idOfCheckbox = 0
+			const markdownLines = this.value.split('\n')
+			markdownLines.forEach((line, i) => {
+				// Regex Source: https://github.com/linsir/markdown-it-task-checkbox/blob/master/index.js#L121
+				// plus the '- '-string.
+				if (/^(- )\[[xX \u00A0]\][ \u00A0]/.test(line.trim())) {
```
korelstar, line 78:
```suggestion
				if (/^-\s+\[[xX \u00A0]\][ \u00A0]/.test(line.trim())) {
```
This did not work correctly if there are multiple spaces after the dash, e.g. for `-  [x] item`

src/components/EditorMarkdownIt.vue
```diff
@@ -51,7 +55,49 @@ export default {
...
+		checkLine(line, index, id, target) {
+			let returnValue = line
+			if ('cbx_' + id === target.id) {
+				if (target.checked) {
+					returnValue = returnValue.replace('[ ]', '[x]')
+					returnValue = returnValue.replace('[\u00A0]', '[x]')
+				} else {
+					returnValue = returnValue.replace('[x]', '[ ]')
+					returnValue = returnValue.replace('[X]', '[ ]')
+				}
```
korelstar, line 96: This does not work correctly if there is a `[ ]` or `[x]` in the item's text, e.g. `- [x] this is an example [x] with checkbox`. In this case, the text is also changed. In order to fix this, it must be assured, that only a single replace is done.

newhinton: As far as i know the .replace() function only replaces the first occurence, when no regex is used. Even if a regex is applied, when the global-flag is missing it will also only chamge the first occurence. So this should be fine, did your test yield different results?

korelstar: Please try my example. It doesn't work correctly. But I did not found out why.

newhinton: It works just fine for me, have you cleared the cache? (I had problems when the developer console wasnt open, it was using some broken hybrid version from cache)

newhinton: It only has the behaviour you describe if we have both an uppercase and a lowercase X in the line. I will have to fix this particular issue though.

korelstar: I've found out that it happens on sub-items only, see the following example:
 ```
- [ ] check [ ] box (works)
  - [ ] sub check [ ] box (does not work)
- [ ] check [ ] box (works)
```

**Comment by korelstar** [comment]
> So, i fixed the single click only bug. Sadly i had to introduce a timeout function so that the v-html can render and the new content is actually found with the checkboxes. If you have a better idea please tell me.

I think it's fine, for now.

> Also i found a very strange bug, but i think this is the issue with MarkdownIT.
> 
> ```
>  - [x] test
>  - [x] test2
>  [ ] test
>  
>  
>  test
> ```
> 
> This note will have 2 checkboxes, as expected, but the label of the second one extends to a second line. Can you confirm this behaviour on your end?

Yes, I can reproduce this. But I agree that this is caused by `markdown-it`.

File: src/components/EditorMarkdownIt.vue [hunk 5]
```diff
@@ -29,7 +33,7 @@ export default {
 		})
 
 		md.use(require('markdown-it-task-checkbox'), {
-			disabled: true,
+			disabled: false,
```
korelstar, line 36:
```suggestion
			disabled: this.readonly,
```

korelstar (changes requested):
> How do i test the read-only attribute? I have passed it just like the editor so it 'should' be fine, but i have never encountered a read only note. Are those shared from other people?

Yes, that's the typical use case. But you could test this also by manually removing write permission to the note's file directly on the file system.

I've added a code change suggestion in order to improve this case: The checkboxes should be disabled if and only if the note is read-only.

src/components/EditorMarkdownIt.vue
newhinton: > - [ ] check [ ] box (works)
>  - [ ] sub check [ ] box (does not work)
> - [ ] check [ ] box (works)


I also cannot reproduce this. Is the sub-element supposed to be indented? Because on my build it behaves identical to the other two (including the checkbox behaviour)

src/components/EditorMarkdownIt.vue
newhinton: It seems this has to be inverted, but i have added this!

korelstar: No, it must not be inverted. With your change, it doesn't work anymore :wink:

src/components/EditorMarkdownIt.vue
korelstar: Yes, the second item should be indented. See this screenshot:

Screenshot

src/components/EditorMarkdownIt.vue
newhinton: Yeah, you're right. Having like 20 notes all called "new note" does not help distinguising which one was shared and which ones were not. I was wondering why i had to invert it :facepalm:

korelstar:
@newhinton What's the state of this PR? As far as I can see, there is only one bug left: the issue with the sub-items (see https://github.com/nextcloud/notes/pull/861#discussion_r917435636).  Any ideas on how to fix this?

korelstar:
> @newhinton What's the state of this PR? As far as I can see, there is only one bug left: the issue with the sub-items (see [#861 (comment)](https://github.com/nextcloud/notes/pull/861#discussion_r917435636)). Any ideas on how to fix this?

Okay, I found a fix. It was because clicking on a sub-item triggers an event for the sub-item AND for the parent item. I had to stop event propagation.

korelstar (approved):

PR #873, opened 2022-07-09
Add a button to create a new note to the dashboard

### LINKED ISSUE #869: Add a "New note" button to the dashboard widget
Right now, there doesn't seem to be a way to create notes or get to the notes app from the dashboard

Just like how the Tasks app has a "New Task" button, the notes app should have a button to quickly open the app to make a new note.

Bookmarking is always an option. But that would add extra clutter to my bookmarks menu

salixor:
Closes #869 by adding a new button :

| No notes | Some notes | A lot of notes |
|--------------|-----------|------------|
| image | image| image |

For reference, next to the Tasks app widget :

image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -213,13 +213,10 @@ export default {
 			}
 		},
 
-		routeToNote(noteId, query) {
-			if (this.$route.name !== 'note' || this.$route.params.noteId !== noteId.toString()) {
-				this.$router.push({
-					name: 'note',
-					params: { noteId: noteId.toString() },
-					query,
-				})
+		routeToNote(id) {
+			const noteId = id.toString()
+			if (this.$route.name !== 'note' || this.$route.params.noteId !== noteId) {
+				this.$router.push({ name: 'note', params: { noteId } })
 			}
 		},
 
@@ -228,13 +225,13 @@ export default {
 				return
 			}
 			this.loading.create = true
-			createNote(this.filter.category || '')
+			createNote(this.filter.category)
 				.then(note => {
-					this.routeToNote(note.id, { new: null })
+					this.routeToNote(note.id)
 				})
 				.catch(() => {
 				})
-				.then(() => {
+				.finally(() => {
 					this.loading.create = false
 				})
 		},
--- a/src/App.vue
+++ b/src/App.vue
@@ -213,10 +213,14 @@ export default {
 			}
 		},
 
-		routeToNote(id) {
+		routeToNote(id, query) {
 			const noteId = id.toString()
 			if (this.$route.name !== 'note' || this.$route.params.noteId !== noteId) {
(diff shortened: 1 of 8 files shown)


src/App.vue
```diff
@@ -228,13 +225,13 @@ export default {
 				return
 			}
 			this.loading.create = true
-			createNote(this.filter.category || '')
+			createNote(this.filter.category)
 				.then(note => {
-					this.routeToNote(note.id, { new: null })
+					this.routeToNote(note.id)
```
korelstar, line 230: Why do you remove the parameter `new`? It has the purpose, that the note's title is automatically generated from the first line of the note.

src/components/Dashboard.vue
```diff
@@ -71,9 +79,14 @@ export default {
...
 			return (note) => {
-				return generateUrl('/apps/notes/note/' + note.id)
+				return generateUrl(`/apps/notes/note/${note.id}`)
 			}
 		},
+
+		buttonsFooterStyle() {
+			const marginTop = this.items.length > 0 ? 20 + (this.displayedItemsCount - this.items.length) * 60 : 10
+			return { marginTop: `${marginTop}px` }
+		}
```
korelstar, line 89: This looks quite complicated. You are comparing to the tasks app, but I don't see such a complicated calculation, there. Do we really need this or can we copy the styling from the tasks app?

korelstar (commented):
Thanks for your contribution, @salixor ! This is a good solution. However, I have some questions to the code changes (see inline comments).

src/components/Dashboard.vue
salixor: The main motive behind this implementation was to have the button stay at the bottom of the widget, rather than below the content. In Tasks, the button stays below the content :

| | |
|--------------|-----------|
| Tasks Widget | Notes Widget |

I'd argue having the button at a fixed height is better, but we could as well simplify the logic and have the button stay below the content.

src/App.vue
salixor: My bad, I hadn't seen it was used somewhere else ! I'll remove these changes

korelstar (changes requested):
I suggest to do absolute positioning instead of calculating the position from top.

salixor:
> I suggest to do absolute positioning instead of calculating the position from top. This also obsoletes `displayedItemsCount`. What do you think?

I had tried it, but I didn't reach a satisfying solution. I'm now realizing I simply had put the `relative` attribute on the wrong `div` 🤦 

Thanks for the suggestion, it's definitely cleaner this way 👍

I also replaced the translation string to prevent the need of translating to every language, since there's already a string that does the job well

korelstar (approved):
Thanks! Looks good and sorry for the delay

PR #878, opened 2022-07-10
Update table style

newhinton:

This uses a more modern table layout, inspired by the collectives app.

Also thanks to [Dai](https://stackoverflow.com/a/2586780) on Stackoverflow for the table-workaround!

### DIFF
--- a/src/components/EditorMarkdownIt.vue
+++ b/src/components/EditorMarkdownIt.vue
@@ -231,7 +231,7 @@ export default {
 	}
 
 	& table td, & table th {
-		padding: 0.75em;
+		padding: 0.35em 0.5em;
 		text-align: left;
 		border: 1px solid var(--color-border);
 	}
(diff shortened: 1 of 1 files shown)


korelstar:
Could you please add some screenshots? (Before and after)

newhinton:
| Original  | New |
| ------------- | ------------- |
| grafik | grafik |

The green tint is the hover-effect, similar to how text/collectives does it.

korelstar (changes requested):
I really like your PR, the screenshot is looking good! However, I think your last commit was broken. Now, it looks like this:

Screenshot

Another thing: we should not use pixel units here. What about my suggested change?

newhinton:
@korelstar sorry, what suggestion?

korelstar:
> @korelstar sorry, what suggestion?

Sorry, my suggestion was lost (don't know why).

> Edit: My tests revealed that 0.75 em as padding works good.

This is too much from my perspective. I did a little adjustment and I think it's ready to merge, now.

PR #883, opened 2022-07-21
The ETag validation fix

jsixface:
The validation was failing if the client is sending the same ETag for a note in the server. This fixes it.


stefan-niedermann:
I disgree. The `"` marks are *part* of the `ETag` (which is *not* per definition a hash but can in fact be anything). So while a quoted `ETag` might be uncommon it is still valid.

The PR introduces a breaking change of the API and requires each client to adapt the changes.

I vote for keeping it the way as it is. A clear hint / description in the API documentation makes sense though.

jsixface:
I updated to strip off the quotes before comparing, so that it will work regardless of the client sending with or without qutoes.

korelstar:
Sorry, I don't understand the problem. When requesting a note, the app add's the `ETag` response header, which contains the `"`, e.g.:
```
ETag: "18411b7505549c97d23c64e070cae388"
```

When updating a note, the client should send the `If-Match` header exactly like before, i.e. with `"`. E.g.:
```
If-Match: "18411b7505549c97d23c64e070cae388"
```

This is exactly what the HTTP standard requires, see https://httpwg.org/specs/rfc7232.html#header.etag

So, where is the problem? Why do we need this PR?

jsixface:
I didn't realise that ETag is a special header and supposed to have quotes. I was under the assumption that it is just another header and various clients may or may not send all headers with quotes. Closing this PR.

PR #892, opened 2022-08-06
make code-blocks nicer

newhinton:

Old Style:
grafik

New Style:
grafik

### DIFF
--- a/src/components/EditorMarkdownIt.vue
+++ b/src/components/EditorMarkdownIt.vue
@@ -293,5 +293,10 @@ export default {
 	& table td:empty::after {
 		content: '\00a0';
 	}
+
+	pre, code {
+		border-radius: 10px !important;
+		padding: 10px !important;
+	}
 }
 </style>
--- a/src/components/EditorMarkdownIt.vue
+++ b/src/components/EditorMarkdownIt.vue
@@ -297,8 +297,8 @@ export default {
 	}
 
 	pre {
-		border-radius: 10px !important;
-		padding: 15px !important;
+		border-radius: 10px;
+		padding: 15px;
 		background: var(--color-background-dark);
 		font-size: 90%;
 		white-space: pre-wrap;
--- a/src/components/EditorMarkdownIt.vue
+++ b/src/components/EditorMarkdownIt.vue
@@ -149,7 +149,7 @@ export default {
 		setSingleCodeRule() {
 			this.md.renderer.rules.code_inline = function(tokens, idx, options, env, self) {
 				const token = tokens[idx]
-				return '<p class="singleLineCode"><code>' + token.content + '</code></p>'
+				return '<code class="singleLineCode">' + token.content + '</code>'
 			}
 		},
 	},
@@ -211,12 +211,6 @@ export default {
 
 	& a {
 		color: var(--color-primary-element);
-	}
-
-	& pre, & code {
-		background: var(--color-background-dark);
-		font-size: 90%;
-		white-space: pre-wrap;
 	}
 
 	& pre code {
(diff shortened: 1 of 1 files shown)


stefan-niedermann:
Nice!

1. How does the `singular line block` play out when there is more text in the line?
2. Is the leading <kbd>Space</kbd> in the "new" sample intented? Where does it come from? Or has it wrongly been stripped in the "old" sample?

newhinton:
> 1. How does the singular line block play out when there is more text in the line?

Oh yeah, it breaks in a funny way. I will have to see if this can be fixed easily.

> 2. Is the leading Space in the "new" sample intented? Where does it come from? Or has it wrongly been stripped in the "old" sample?
That is a bug. I guess it is due to the padding applied to give the content a bit more space, but it seems it is only applied to the first line.

stefan-niedermann:
> That is a bug. I guess it is due to the padding applied to give the content a bit more space, but it seems it is only applied to the first line.

Without having a look into it, the behaviour hints for a CSS `text-indent` property. Maybe there is one which needs to be overridden :shrug:

newhinton:
grafik

stefan-niedermann:
Looks good :) Just for clarification: This is only for code *blocks* (three backticks), and *not* for *inline* code (one backtick), right?

newhinton:
Both! Block 1&2 in the screenshot are single ticks, 3 is block.

But now that you say it, let me check how well that behaves inside a larger text. Maybe i need to not apply a padding there

newhinton:
grafik

I fixed the issue for textblocks with quotes, but multiline-singleline-blocks look a bit worse than before. I dont think that is fixable, but maybe that is a tradeoff

src/components/EditorMarkdownIt.vue
```diff
@@ -143,6 +145,13 @@ export default {
 				}
 			}
 		},
+
+		setSingleCodeRule() {
+			this.md.renderer.rules.code_inline = function(tokens, idx, options, env, self) {
+				const token = tokens[idx]
+				return '<p class="singleLineCode"><code>' + token.content + '</code></p>'
```
stefan-niedermann, line 152: I don't know much about the development process and used transpilers, but
1. is it possible to use javascript template strings for a more readable code?
2. do we need to check whether `token.content` is available? something like `token?.content` or similar?

src/components/EditorMarkdownIt.vue
```diff
@@ -143,6 +145,13 @@ export default {
 				}
 			}
 		},
+
+		setSingleCodeRule() {
+			this.md.renderer.rules.code_inline = function(tokens, idx, options, env, self) {
```
stefan-niedermann, line 150: Is there any reason we don't use (IMHO) more readable arrow functions? Some Nextcloud codinf style guide or some transpiler limitation?

src/components/EditorMarkdownIt.vue
```diff
@@ -143,6 +145,13 @@ export default {
 				}
 			}
 		},
+
+		setSingleCodeRule() {
```
stefan-niedermann, line 149: I am confised by the term `singleCode` (as well as the css class pendant). Does it mean
- code block (triple backticks) without a line break or
- an inline code term (single backticks)
- both?

src/components/EditorMarkdownIt.vue
newhinton: 1. Until now, i did not know that javascript HAD template strings, thats a nice find :D
2. I dont think so. When i understood the MarkdownIT library correctly, token can never be undefined. content may be, but that would seem like a bug, but i will improve null-safety!

src/components/EditorMarkdownIt.vue
newhinton: I did that because MarkdownIT had an example that i copied, but no technical reason. I will experiment with changing that!

src/components/EditorMarkdownIt.vue
newhinton: Naming is not my strong suit, i'll change that with the other stuff

korelstar:
@newhinton Thanks for your PR, this looks good! And thanks @stefan-niedermann for the review!

@newhinton What's the state of this PR? Are all issues fixed? Is this ready for final review? Could you please update the screenshots?

newhinton:
@korelstar Yes this is ready! 

image

korelstar:
Looks very nice! Will add this to the 4.6.0 release, after the `pre-wrap` is fixed :tada:

PR #908, opened 2022-08-29
[wip] implement client dashboard api for notes

icewind1991:
Dashboard api will allow clients to render the dashboard widgets

See https://github.com/nextcloud/server/pull/33658 for more information, requires the interfaces added by that pr

This implementation was created to provide an example for a dashboard that has a button. Interfaces might still be subject to change.

### DIFF
--- a/lib/AppInfo/DashboardWidget.php
+++ b/lib/AppInfo/DashboardWidget.php
@@ -4,20 +4,28 @@ declare(strict_types=1);
 
 namespace OCA\Notes\AppInfo;
 
+use OCA\Notes\Service\Note;
+use OCA\Notes\Service\NotesService;
+use OCP\Dashboard\IAPIWidget;
+use OCP\Dashboard\IButtonWidget;
 use OCP\Dashboard\IWidget;
+use OCP\Dashboard\Model\WidgetItem;
 use OCP\IL10N;
 use OCP\IURLGenerator;
 
-class DashboardWidget implements IWidget {
+class DashboardWidget implements IWidget, IButtonWidget, IAPIWidget {
 	private IURLGenerator $url;
 	private IL10N $l10n;
+	private NotesService $notesService;
 
 	public function __construct(
 		IURLGenerator $url,
-		IL10N $l10n
+		IL10N $l10n,
+		NotesService $notesService
 	) {
 		$this->url = $url;
 		$this->l10n = $l10n;
+		$this->notesService = $notesService;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


lib/AppInfo/DashboardWidget.php
```diff
@@ -61,4 +69,31 @@ public function getUrl(): ?string {
...
+	public function getButtonUrl(): string {
+		return $this->url->linkToRouteAbsolute('notes.page.create');
+	}
+
+	public function getButtonIconUrl(): ?string {
+		return null;
+	}
+
+	public function getButtonText(): string {
+		return $this->l10n->t('New notes');
```
stefan-niedermann, line 82: Shouldn't this be singular? <kbd>New note</kbd>?

lib/AppInfo/DashboardWidget.php
```diff
@@ -61,4 +70,29 @@ public function getUrl(): ?string {
...
+	public function getItems(string $userId, ?string $since = null, int $limit = 7): array {
+		$notes = $this->notesService->getTopNotes($userId);
+		$notes = array_slice($notes, 0, $limit);
+		return array_map(function (Note $note) {
+			$excerpt = '';
+			try {
+				$excerpt = $note->getExcerpt();
+			} catch (\Throwable $e) {
+			}
+			$link = $this->url->linkToRouteAbsolute('notes.page.index', ['id' => $note->getId()]);
```
tobiasKaminsky, line 93: Is this correct link?
It is something like: https://cloud.com/index.php/apps/notes/?id=271
but on web I see:
https://cloud.com/apps/notes/note/6004520

Thus all links via widget end up showing any previous note but not the desired one.

korelstar: Yep, it must be `notes.page.indexnote` instead:

```suggestion
			$link = $this->url->linkToRouteAbsolute('notes.page.indexnote', ['id' => $note->getId()]);
```

See e.g. https://github.com/nextcloud/notes/blob/v4.5.0/lib/Controller/PageController.php#L62

korelstar:
Nice! How can I test this (i.e. which clients will use that API)? Is it for Nextcloud Android/iOS or for the Nextcloud server itself? The last one would be interesting, too, because currently, the dashboard app loads several megabytes Javascript for every widget.

tobiasKaminsky:
It can be used by any client, but first we will implement it on Android/iOS.

tobiasKaminsky:
Bug: iconURL is empty although it is shown in web @robi

tobiasKaminsky:
Bug: notes icon url on widgets-items endpoint is empty

juliushaertl:
@tobiasKaminsky I rebased and pushed a commit to fix the icon bugs. Mind to give this another test run?

tobiasKaminsky:
In {{ base_url  }}/ocs/v2.php/apps/dashboard/api/v1/widgets:
icon_url: "\/nc\/apps\/notes\/img\/notes-dark.svg"

but it must be absolute, otherwise clients cannot use it.


In: {{ base_url  }}/ocs/v2.php/apps/dashboard/api/v1/widget-items
iconURL: is starred.svg or notes-dark.svg
But it should provide preview of text.

image

Issue #907, opened 2022-08-29
Bug when deleting a note


Kitanit:
next 24.0.3
latest version of notes

![image](https://user-images.githubusercontent.com/65107731/187147415-ab2a766d-e0b8-4095-ad3c-3e581f71c7be.png)

stefan-niedermann:
Indeed, looks like a typo (`tilte`), so the property binding fails.

korelstar:
Thanks for reporting this. This app uses [Transifex](https://www.transifex.com/nextcloud/nextcloud/notes/) for translations. Hence, this error has to be fixed in the translation over there. You could create an account over there and help with translations.

But for this specific issue, maybe @rakekniven could also help?

korelstar:
Thanks!

PR #910, opened 2022-09-02
fix russian translation

miles170:
fixes: #907


korelstar:
Thanks for your contribution! However, this app uses [Transifex](https://www.transifex.com/nextcloud/nextcloud/notes/) for translations. There is an automatic process which daily commits all translations from Transifex. Therefore your change will be automatically overridden on the next day. In order to fix this issue, the translation on Transifex has to be fixed.

Issue #918, opened 2022-09-21
Paste/dragdrop images


aPollO2k:
It should be possible to paste/dragdrop images directly into a note without the need of save the image as file on disk and upload manually.

stefan-niedermann:
I agree, maybe you want to start a Pull Request for this? :rocket:

demuxer86:
The only reason I can't use this app is because I can't add pictures to the notes from my android phone.   It's a key feature to any note app imo.    Taking a picture of something, being able to do basic drawings on that picture, and then adding some text to the note.   That's invaluable and Google Keep is the only note app that does it mostly right.

stefan-niedermann:
@demuxer86 your feature request is not related to this issue. The Notes Android app [is waiting for an external API for attachments](https://github.com/nextcloud/notes/issues/826) and will implement support for adding attachments when it is available. Please use GitHub reactions (👍) in the future when you are affected by an issue and have no new information to add. Thank you.

PR #958, opened 2023-01-30
feat(editor): Use text editor

juliushaertl:
- [x] 🐛 Check issues with sidebar loading when using the new editor
- [x] Add option to switch to the new editor
- [x] Make it default for new installations
  - [x] Ensure .md is the default file format
  - [x] Ensure that the text editor is the default but fallback to easymde if text is unavailable
- [x] Migration process (if notes was installed before)
  - [x] Show information about new editor and suggest to switch to it

## From the old pull request
- [ ] remember last opened note
- [x] auto title for new notes
- [ ] don't break original file, see https://github.com/nextcloud/text/issues/593 and https://github.com/nextcloud/text/issues/772
- [x] improve parallel changes from outside of app, see https://github.com/nextcloud/text/issues/1212
- [x] use correct / other list/task syntax, see https://github.com/nextcloud/text/issues/1297
- [x] loading another note with same component (proof-of-concept uses a "hack" which does not allow for a loading indicator)
- [ ] "Link file" feature of the Text app could be replaced by a "Link note" feature (?)
- [ ] how to provide integrated images to 3rd-party apps using the Notes API?

## List of issues this addresses/resolves

- Fixes https://github.com/nextcloud/notes/issues/331
- Fixes https://github.com/nextcloud/notes/issues/918


newhinton:
> how to provide integrated images to 3rd-party apps using the Notes API?

this is already implemented, see here: #785

newhinton:
> Ensure .md is the default file format

We probably should think very careful about that one. Otherwise we have the same issue as in https://github.com/nextcloud/text/issues/593.

Why dont we do the following:

- The user did _not_ change the default file-ending:
   - Use `.ncdata` if nextcloud/text is used
   - Use `.md/.txt` if easymde is used (depending on if we want to stick to the current default or change it)

- The user _did_ change the default file-ending:
   - Use what the user supplied
   
This way we dont get into that whole discussion again, while keeping it relatively simple.

The UX would be that the dropdown for the editor would also change the displayed value for the file-ending. That would not reqire any additional ui-elements.

The additional benefit is that those files would also be editable without the notes app, directly from the files app.

szaimen:
Is master not protected?

juliushaertl:
Right, added that.

julien-nc (commented):
Didn't read the code yet.

1. I face the conflict view after editing in "rich text mode", not waiting for the top-right checkmark to say it was saved, switching to "edit mode" and making a change.
2. After a change in "edit mode", even after the document has been saved (asterisk disappears in the navigation menu), switching to "rich text mode" also brings up the conflict view.

Maybe those issues are not so important as few people will switch between display modes...

Otherwise works like a charm.

max-nextcloud (approved):
I had a look at the code only. Looks good to me.

juliushaertl:
Let's handle the conflict when switching separately. We need some extra handling for ensuring that the file is saved before switching views then.

PR #959, opened 2023-01-31
Enable admins to set the settings defaults

### LINKED ISSUE #629: extension .md as default for every new user
Open notes files on pc - opens it in normal text viewer as normal text because of the extension .txt (no markdown format)

Option to set .md for markdown as default extension for all new user

notify every new user to manually set .md as default

**Additional context**
I know the setting in option in the bottom left of the Web UI, but it is only for one (current) user

h0lley:
I needed to be able to customize the defaults of Note settings (i.e. name of folder to store notes in, file extension, and display mode).

Mainly because in my case, it's poor UX for users to jump directly into edit mode when browsing notes (articles), and `.md` always makes more sense than `.txt`.

It can now be done using `occ` like this:

```
occ config:app:set notes noteMode --value="preview"
occ config:app:set notes fileSuffix --value=".md"
occ config:app:set notes defaultFolder --value="Wiki"
```

a GUI in admin settings may be added later.

fixes #629

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,3 +43,20 @@ Before reporting bugs:
 ### REST API for third-party apps
 
 The notes app provides a JSON-API for third-party apps. Please have a look at our **[API documentation](docs/api/README.md)**.
+
+
+### Admin configuration
+
+It is possible to specify different defaults for the notes settings of new users using `occ` commands like these:
+
+```
+occ config:app:set notes noteMode --value="preview"
+occ config:app:set notes fileSuffix --value=".md"
+occ config:app:set notes defaultFolder --value="Shared notes"
+```
+
+| Setting | Property name | Default | Other available option(s) |
+|---------|---------------|---------|---------------------------|
+| Display mode for notes | noteMode | .edit | .preview |
+| File extension for new notes | fileSuffix | .txt | .md |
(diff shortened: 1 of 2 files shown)


stefan-niedermann:
> and .md always makes more sense than .txt.

IIRC it was set to `.txt` by default because otherwise windows users will unable to open a synced file via double click with notepad when they don't have installed a proper markdown editor.

Maybe notepad can handle md files with win 10 or 11?

h0lley:
I think there's still no pre-installed application associated with the `.md` extension under win 10.

it's kind of unrelated though, no? note that my comment pertains to "in my case". :)
we'll either be using the web interface or will coordinate in a small team which software to use for `.md` files.

another part of my particular use-case is that I'd like to be able to use a globally shared folder (with a name other than `Notes`) as default for new users.

juliushaertl:
Thanks for the pull request, the changes itself look good. Could you maybe add a short documentation section for this to the end of the README as well?

juliushaertl (approved):

h0lley:
sure, done. :)

want me to squash to one commit?

juliushaertl:
That would be much appreciated :)

PR #964, opened 2023-02-08
perf: Lazy load dashboard components

juliushaertl:
Save loading and evaluating larger amounts of javascript on the early page load.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,4 +3,14 @@ const path = require('path')
 
 webpackConfig.entry.dashboard = path.join(__dirname, 'src', 'dashboard.js')
 
+webpackConfig.optimization = {
+	splitChunks: {
+		cacheGroups: {
+			defaultVendors: {
+				name: 'commons',
+			}
+		},
+	},
+}
+
 module.exports = webpackConfig
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -1,10 +1,11 @@
-import Vue from 'vue'
-import Dashboard from './components/Dashboard.vue'
-
-Vue.mixin({ methods: { t, n } })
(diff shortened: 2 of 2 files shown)


juliushaertl:
@julien-nc Might be something nice for all the integration widgets at some point :)

julien-nc:
:+1: 
It does not work with a subpath on my local setup. I see a request to `https://nextcloud.local/apps/notes/js/notes-commons.js?v=1fe6d2ef17a17417d45d` while it should be `https://nextcloud.local/dev/server26/apps/notes/js/notes-commons.js?v=1fe6d2ef17a17417d45d`

webpack.config.js
```diff
@@ -3,4 +3,14 @@ const path = require('path')
 
 webpackConfig.entry.dashboard = path.join(__dirname, 'src', 'dashboard.js')
 
+webpackConfig.optimization = {
+	splitChunks: {
+		cacheGroups: {
+			defaultVendors: {
+				name: 'commons',
```
julien-nc, line 10:
```suggestion
				name: 'dashboard',
```
This fixed it, at least in my context. Was it a mistake or did you intentionally wrote "commons" here?

julien-nc (changes requested):

webpack.config.js
julien-nc: Looking a bit at the Webpack doc and other apps, maybe this is the proper fix:
```
__webpack_nonce__ = btoa(OC.requestToken) // eslint-disable-line
__webpack_public_path__ = OC.linkTo('notes', 'js/') // eslint-disable-line
```

I don't quite get how one chooses what to put in which chunk file.

juliushaertl: Actually the second one would be the proper fix, sorry I missed that.

julien-nc (approved):
:+1:

Issue #960, opened 2023-02-01
Change icon


tobiasKaminsky:
Jan and me had a design review on Android and it was decided to slightly update the icon.
Please therefore also use the same on web: https://fonts.google.com/icons?selected=Material+Icons:edit:

marinofaggiana:
Preview (draft):

<img width="95" alt="190-196-max" src="https://user-images.githubusercontent.com/8616947/220882774-65422c56-da75-4d60-b218-dc4b1da8fb82.png">

AlvaroBrey:
> Preview (draft):
> 
> <img alt="190-196-max" width="95" src="https://user-images.githubusercontent.com/8616947/220882774-65422c56-da75-4d60-b218-dc4b1da8fb82.png">

That's the old pencil icon

PR #968, opened 2023-02-23
enh: Update icons to material design

juliushaertl:
Fixes https://github.com/nextcloud/notes/issues/960


jancborchardt (changes requested):
Nice! :) Only `img/favicon-touch.png` seems wrong, should be same as the `.svg`, with the blue background.

Issue #991, opened 2023-03-23
Contrast in help is bad in dark mode


Wv5twkFEKh54vo4tta9yu7dHa3:
## Steps to reproduce
1. Enable dark mode
2. go to Notes app
1. Scroll down notes list
3. press help button
4. scroll down to shortcuts 

### Expected behaviour
Shortcuts list should be readable

### Actual behaviour
It is not readable

### Screenshots
![swappy-20230323_095336](https://user-images.githubusercontent.com/61561059/227151823-af2ccc74-41d4-4a4c-a57d-826e2bd0c443.png)

## Server

- Notes app version: 4.7.2

- Nextcloud version: 25.0.4

## Client

- Browser (incl. version): Firefox 110

eckelon:
Hi! I saw this issue and I think I fixed it, it was just adding a color property in a the `table.notes-help tr:nth-child(even) {
` query in `AppHelp.vue`
I made it black but I'm not sure if that should be the right color. I'm submitting a PR if it's OK.

Thanks!
<img width="773" alt="image" src="https://user-images.githubusercontent.com/11048462/227266878-20632d10-5f1c-424d-8ebf-278e73abb58b.png">

PR #1000, opened 2023-03-23
Fix help table on dark mode

eckelon:
This PR tries to fix https://github.com/nextcloud/notes/issues/991

This is my first PR so thank you for your patience for reviewing it.

### DIFF
--- a/src/components/AppHelp.vue
+++ b/src/components/AppHelp.vue
@@ -182,7 +182,7 @@ table.notes-help td:first-child, table.notes-help th:first-child {
 }
 
 table.notes-help tr:nth-child(even) {
-	background-color: #eeeeee;
+	background-color: var(--color-background-dark);
 }
 
 </style>


src/components/AppHelp.vue
```diff
@@ -183,6 +183,7 @@ table.notes-help td:first-child, table.notes-help th:first-child {
 
 table.notes-help tr:nth-child(even) {
 	background-color: #eeeeee;
+	color: black;
```
juliushaertl, line 186: Thanks a lot for your pull request. I think the better approach would be to adjust the background color to make sure it uses the css variable we have in place at Nextcloud which adapts to the dark/light mode dynamically.

```suggestion
	background-color: var(--color-background-dark);
```

eckelon: oh, that makes sense! I tried it and it works fine on my machine (images attached). I updated the PR.

Huge thanks!

image
image

juliushaertl:
Thanks a lot for the fix :)

Issue #967, opened 2023-02-21
Design review notes


juliushaertl:
From the design review call today:

Combined review of #842 and #958 

# Notes design review

21\.02.2023: Julius, Jan, Nimisha, Florian, Joas, Maksim, Jonas

## Overall

* [x] Looks really nice! :) And the 3-column layout is also cool

## Accessibility

* [ ] #971

## Navigation

* [ ] #1161
* [x] Between "Uncategorized" and the category list there should be a header "Categories", like in Contacts
* [ ] https://github.com/nextcloud/notes/pull/1002 Also like in Contacts, there could be a "+" to quickly add categories directly there
  * [ ] Then later we can talk about a way to quickly remove categories via there. Currently the category is removed if there is no note in it. 2 options for deletion could be to 1) Delete category only, notes will be uncategorized then or 2) Delete category including the notes in it (like a folder deletion).

## List view

* [x] More spacing around the icon in the list
* [x] Have same action menu for notes in the list view as in the notes detail view
* [x] Entries need margin to left and right so they don’t touch the navigation on the left and the border on the right
* [x] Adjust style of the date headings
  * [x] More spacing to the top
  * [x] Make sure it uses same component as e.g. in the Talk search for category dividers
* [ ] Within categories, sorting should also just be by most recently edited (within subcategories as well)

## Detail view

* [x] Breadcrumbs bar is not needed, the category list in the navigation is enough
* [x] 3-dot menu is primary style, should be standard icon-only
* [x] Title handling: Adjust more to Apple Notes in that the first line is taken as title, also if you change it manually

## Sharing/details

* [x] https://github.com/nextcloud/notes/issues/285
  * Sharing is not implemented in Notes yet, we would need to pull in the sharing sidebar somehow → something for after 26

## Sidebar

* [x] #970
* [x] Could be removed since the character / word count is in the 3-dot menu of the formatting bar

## Settings

* [x] #969

## Help

* [ ] #1162
* [x] Just keep the content that is in the emptycontent, showing the brief help and the links to the mobile apps

juliushaertl:
Updated checkboxes with the current state of #1003 and https://github.com/nextcloud/notes/pull/1021 being merged.

juliushaertl:
Closing as larger topics are done and smaller leftovers are moved to separate issues.

PR #1002, opened 2023-03-24
Category: Add new category button to create new folders

theatischbein:
Following #967 I added an action in the navigation item `Categories` for adding new folders in the current notes path acting as categories.

Some remarks:
- I used the `remote.php` api end-point like the files app, but I am unsure if there is a more elegant way https://github.com/nextcloud/server/blob/55ba9601f2ef5b14524693a77c59cb7f7174270c/apps/files/src/utils/davUtils.js
- I wanted to use `NcActionInput` like the calendar app, but I had trouble to get the trailing submit arrow working, because it only works if the value isn't empty at rendering.. https://github.com/nextcloud/nextcloud-vue/blob/c269d6a3c255f6c62cdefe9407f3756911ee8964/src/components/NcActionInput/NcActionInput.vue#L221
- The Category list does not display empty folder, after creating a category I seleted it so that the success is visible, but if no note is added, the category will vanished from the list until a note is added this category
- The category list will count wills in folder even they are not matching the select file suffix
- Opening the add menu is possible without expanding the category list (just shown in gif differently)

This gifs show the workflow and an error case, if a folder with that name already exist (but may not listed in categories if the folder is empty)

ezgif com-video-to-gif

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "4.8.1",
       "license": "agpl",
       "dependencies": {
+        "@nextcloud/auth": "^2.2.1",
         "@nextcloud/axios": "^2.4.0",
         "@nextcloud/dialogs": "^4.2.1",
         "@nextcloud/event-bus": "^3.1.0",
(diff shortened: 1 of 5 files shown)


juliushaertl:
Just noticed that showing all categories is at least an option as part of the 3-column layout in https://github.com/nextcloud/notes/pull/842/commits/4412195062b3b9e9a12100941fbbcc3c2276a582

jancborchardt:
> Just noticed that showing all categories is at least an option as part of the 3-column layout in https://github.com/nextcloud/notes/commit/4412195062b3b9e9a12100941fbbcc3c2276a582

Yeah, we could simply also show categories without notes in them, that is fine. If the folder exists it should be shown. :)

And design-wise yep, this depends a bunch on the 3-column layout. With the switch to that, we don’t need this whole expand and collapse mechanic anymore, as the folder/category list is the only thing in the left navigation.

JonnyTischbein:
Another ping here @juliushaertl does this need to await the redesign to the 3 column layout ?

Are they any open tasks that I can implemend here ?

juliushaertl:
We merged the three column layout end of last week, i think this change would be a great addition to rebase on top.

Regarding other tasks it mostly depends on your interest, we are generally open to any feature contribution, if there is an issue if of course might make sense to discuss ideas first to also align with our designers about potential UX topics.

We also have a chatroom for contributors on our Nextcloud instance, in case you are interested in joining for some exchange feel free to drop me an email to `julius at nextcloud.com` then I'd be glad to send you an invite.

juliushaertl:
I also gave this a rebase and during that adjusted the positioning of the new category button which could now be located right next to the section header:

Screenshot 2023-11-07 at 15 49 39

The main missing part would be to properly handle displaying the new category in the list now.

theatischbein:
Hi @enjeck,
thanks for requesting to take over! 
I totally forgot about this PR and do not have the time to finish it in the next time.
So I would be very happy if you continue on your own base. But feel free to use as many things from this PR!

jancborchardt (commented):
Yep, still would be nice. :) @enjeck if you need any design input let @marcoambrosini and me know, but it looks quite fine already.

(If needed/wanted we can also switch over to do it via a modal, but to be honest the inline way of creating is more elegant: You click a button on the left, and the category will be created on the left. Moving to the center for a modal would be strange FYI @marcoambrosini )

JuliaKirschenheuter:
Dear @theatischbein,

it would be great to merge your PR and go life with a new improvement! Could you please answer Marco's question? Do you need some more support, to finish your improvement?

enjeck:
Thanks, all. I’m closing this PR because it is now very old and the original implementation no longer fits the current Vue 3/Pinia/sidebar code. Current main already has category creation from the left navigation. there's the “New category” button opens an inline editable draft row, adds the new category locally, selects it, and supports moving a dragged note into a newly named category.

Issue #970, opened 2023-02-24
Category adjustment could go in the note menu


juliushaertl:
After that the sidebar could be removed as word count is being added through text #958

PR #1004, opened 2023-03-25
Sidebar: removed

JonnyTischbein:
Following #967 I removed the sidebar, because the only remaining function category can also be accessed via the app navigation. Also the sidebar isn't accessable in the richtext editor anyways.


juliushaertl:
Glad to see so many awesome contributions @JonnyTischbein I'll probably only get to review at the end of this week, but I try to go through them as soon as possible.

juliushaertl:
So currently the app navigation actions menu does not offer a way to set the category just to navigate to it, which is something we'd need to address differently if we drop the sidebar. (https://github.com/nextcloud/notes/issues/970)

We don't have any outline yet how that could look like, so if you would have any suggestion for that UI wise we should discuss that first. Otherwise maybe @nimishavijay or @jancborchardt have more input for how the category could be selected in the navigation.

JonnyTischbein:
> So currently the app navigation actions menu does not offer a way to set the category just to navigate to it, which is something we'd need to address differently if we drop the sidebar. (https://github.com/nextcloud/notes/issues/970)

My suggestion would be to move the Category dropdown menu from the sidebar to `....` context menu from a note.
Maybe in a small modal like adding a new Category in #1002 or the contact `Add Group`

JonnyTischbein:
I am not fully content with this, but didn't know how to trigger a popper div from an action menu.

What do you think ?

image

JonnyTischbein:
I did another mockup, but still keep to figure out how to adjust the appearance of the menu item:
ezgif com-video-to-gif

```
<NcActions :force-menu="true" :force-title="true" :menu-title="actionCategoryText">
    <template #icon>
        <FolderIcon :size="20" />
    </template>				
   <NcSelect id="category"
	:value="category"
	:options="categories"
	:placeholder="t('notes', 'Uncategorized')"
	:disabled="loading.category"
	:class="['category-select', {'icon-loading-small': loading.category}]"
	:show-no-results="false"
	:taggable="true"
        :preserve-search="true"
	:title="t('notes', 'Set category')"
	@input="onSaveCategory"
        @close="onFinishEditCategory"
        @search-change="onEditCategory"
    />
</NcActions>
``

juliushaertl:
Took the liberty to rebase this PR :)

juliushaertl:
I also further pushed a commit to include #1104 and making use of NcActionInput multiselect for switching the category. 

The overflow is not the nicest there yet, but this is something that should probably be fixed upstream.

juliushaertl (approved):

PR #1021, opened 2023-04-13
Three column layout (rebased/squashed)

juliushaertl:
Rebased and squashed version of #842 to have an easier way to resolve the conflicts

So far pushed another commit additionally to bring back the actions menu in the list after my rebase and some minor styling fixes

Screenshot 2023-04-29 at 12 28 00


src/App.vue
```diff
@@ -61,8 +58,8 @@ import PlusIcon from 'vue-material-design-icons/Plus.vue'
 import CogIcon from 'vue-material-design-icons/Cog.vue'
 
 import AppSettings from './components/AppSettings.vue'
-import NavigationList from './components/NavigationList.vue'
 import EditorHint from './components/Modal/EditorHint.vue'
+import CategoriesList from './components/CategoriesList.vue'
```
JonnyTischbein, line 62: I would move it to line 61 to keep the ordering

src/App.vue
```diff
@@ -74,14 +71,14 @@ export default {
...
 		EditorHint,
-		NavigationList,
 		NcAppContent,
 		NcAppNavigation,
 		NcAppNavigationNew,
 		NcAppNavigationItem,
 		NcContent,
 		CogIcon,
 		PlusIcon,
+		CategoriesList,
```
JonnyTischbein, line 81: Again order it alphabetically like in the other files

src/components/CategoriesList.vue
```diff
@@ -1,15 +1,8 @@
...
-		:class="{ 'category-header': selectedCategory !== null }"
-		:open.sync="open"
-		:allow-collapse="true"
-		@click.prevent.stop="onToggleCategories"
-	>
-		<FolderIcon slot="icon" :size="20" />
+	<Fragment>
 		<NcAppNavigationItem
 			:title="t('notes', 'All notes')"
+			:class="{ active: null === selectedCategory }"
```
JonnyTischbein, line 5: Shouldn't it be `selectedCategory === null` ?

src/components/NoteItem.vue
```diff
@@ -27,29 +41,29 @@
...
 export default {
-	name: 'NavigationNoteItem',
+	name: 'NoteItem',
 
 	components: {
 		NcActionButton,
-		NcActionSeparator,
-		NcAppNavigationItem,
+		NcListItem,
+		AlertOctagonIcon,
```
JonnyTischbein, line 64: I would fix the order again

src/components/NoteItem.vue
```diff
@@ -64,50 +78,46 @@ export default {
...
 		},
-
 		actionDeleteIcon() {
 			return 'icon-delete' + (this.loading.delete ? ' loading' : '')
 		},
 	},
 
 	methods: {
+		onNoteSelected(noteId) {
+			console.error('select note', noteId)
```
JonnyTischbein, line 118: Is this a debug log ?

src/components/NotesView.vue
```diff
@@ -0,0 +1,270 @@
...
+		category() {
+			return store.getters.getSelectedCategory()
+		},
+
+		filteredNotes() {
+			return store.getters.getFilteredNotes()
+		},
+
+		displayedNotes() {
+			if (this.filteredNotes.length > 40 && this.showFirstNotesOnly) {
```
JonnyTischbein, line 113: Why you catch for `> 40` but slice to `0-30` ?

src/components/NotesView.vue
```diff
@@ -0,0 +1,270 @@
...
+
+		categoryToLabel(category) {
+			return categoryLabel(category.substring(this.category.length + 1))
+		},
+
+		getTimeslotFromNote(note) {
+			if (note.favorite) {
+				return ''
+			}
+			const t = note.modified * 1000
```
JonnyTischbein, line 186: May move `1000` to const variable

src/components/Sidebar.vue
```diff
@@ -170,6 +195,22 @@ export default {
 				})
 		},
 
+		onUpdateTitle(newTitle) {
+			this.newTitle = newTitle
+		},
+
+		onRenameTitle(event) {
+			if (event && this.title !== this.newTitle) {
```
JonnyTischbein, line 203: Is  it possible that an event is triggered, that is not defined ?
I think it is safe to remove `event` from the if clause

JonnyTischbein (commented):
I will test the function tomorrow, but here some comments on the code

src/components/NotesView.vue
juliushaertl: I'd still keep this as this is a pattern that js developers likely know well and a constant (or timestampToMilliseconds helper that I was thinking about) would be easy to be overseen in new code introduced and then we'd have two ways of doing that.

src/components/NotesView.vue
juliushaertl: Something to consider separately I'd say as it is not introduced here but originating from https://github.com/nextcloud/notes/pull/424/files#diff-84c733529b3b3fd6edfc369f85fe95b5cfb8d1f0a74d39712293aa3b037a0848R113-R117 s

src/components/CategoriesList.vue
juliushaertl: Yep, more what the current code base and Nextcloud apps are using.

juliushaertl:
Thanks for the review @JonnyTischbein, very much appreciated. I addressed most of them, two I'd keep out of this PR.

src/components/Note.vue
```diff
@@ -1,6 +1,6 @@
 <template>
-	<NoteRich v-if="rich" :note-id="noteId" />
-	<NotePlain v-else :note-id="noteId" />
+	<NoteRich v-if="ready && rich" :note-id="noteId" />
```
luka-nextcloud, line 2: I think `v-if="rich"` is enough here because rich already includes condition check of ready (check line 27 and 30).

juliushaertl: Thanks, pushed a bit of a change there to make the state more clear through computed properties

luka-nextcloud (approved):

juliushaertl:
Let's get this in so we can have a first beta for more testing.

juliushaertl:
Hm, I cannot think of any. Can you test with a new user account to see if it might be related to the account settings with which editor is being used as I see NotePlain in the trace?

PR #1031, opened 2023-04-26
fix: Proper response for attachment endpoint

juliushaertl:
Use a stream response to set a Content-Disposition header on the attachments


datenangebot (approved):
Codereview: looks good to me

PR #1047, opened 2023-05-17
fix: Wrap renaming of notes through autotile in locking context

juliushaertl:
Otherwise setting the title of the note will fail when editing through the text editor if files_lock is enabled


blizzz (approved):
looks reasonable, didn't test

juliushaertl:
> Need to double check if setting a category also needs to be adapted, as I got a similar report for that one

Moved the wrapping logic to a helper function and covered the category updates as well, which would run into the lock as they move the file around.

blizzz (approved):

Issue #38, opened 2016-12-16
Share a single note / publish


jancborchardt:
It would be shown in »read only« mode of course (see also https://github.com/nextcloud/notes/issues/23). Basically the page like for share links with the blue header, in the content only the note text, formatted properly (no Markdown highlight stuff) and on the top right a download function.

The share icon could be in a 3-dot menu in the left navigation, where also »Delete« should be placed now that we get more stuff. We could layout it like in the [Spreed app](https://github.com/nextcloud/spreed/), where quick sharing by link is possible:
![capture du 2016-12-15 16-22-04](https://cloud.githubusercontent.com/assets/925062/21229677/b7887026-c2e2-11e6-8444-d500ec5a81ce.png)
Clicking on »Share link« shows the link in an input field below and also shows a copy button. We could also use the clipboard API to directly copy the share link to the clipboard.

We don’t need full-blown sharing with others on the same instance yet I’d say since there’s no realtime collab anyway. This is about publishing mainly. :)

roipoussiere:
Note that you can share a not by sharing the file (in the `Notes` folder) with the File app.

But the note will be stored in the root folder of the user, not in Notes folder.

jancborchardt:
@roipoussiere yep, of course. That requires you to switch the app, find the note, share it, and go back to the Notes app. ;) Hence the direct integration is needed.

palto42:
With the markdown app there is currently also no option to share the rendered markdown (as it was requested here), but there is a related request for such feature https://github.com/icewind1991/files_markdown/issues/29

dcrobertson01:
It needs to be shared in the notes folder not the files folder. This is not sharing a note - it is sharing a file. 

I'd like the shared note to be available in the notes app on the phone.

trurli:
You should be able to make it visible in the notes app by moving the shared file to the Notes folder manually. However this is really annoying.

I looked into the file sharing code and unfortunately the target folder is always set to the system default folder for sharing, usually the user's root folder, so there's no easy way to solve sharing a note directly using the file sharing API unless something changes in the files_sharing app. 

As I'm still fairly new to the whole code-base I'm still unsure how/if dependencies between apps would be handled.

jancborchardt:
@nextcloud/sharing @schiessle any help regarding sharing here is appreciated. :)

ptandler:
The last comment is from 2018 ... any update? Did / will it make it on the roadmap?

[...]

PR #1146, opened 2023-11-01
Feature/share single note

luka-nextcloud:
Resolves: #38 

- Front-end:
  - Add share button to note's menu
  - Add sharing indicator
  - Collapse navigation (left panel) when open sdebar
  - Handle restored event
- Back-end:
  - Dispatch events to load Sidebar script and Viewer script
  - Handle event BeforeShareCreatedEvent to update share target

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -4,11 +4,13 @@ declare(strict_types=1);
 
 namespace OCA\Notes\Controller;
 
+use OCA\Files\Event\LoadSidebar;
 use OCA\Notes\AppInfo\Application;
 use OCA\Notes\Service\NotesService;
 
 use OCA\Notes\Service\SettingsService;
 use OCA\Text\Event\LoadEditor;
+use OCA\Viewer\Event\LoadViewer;
 use OCP\App\IAppManager;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\ContentSecurityPolicy;
@@ -64,6 +66,14 @@ class PageController extends Controller {
 
 		if (\OCP\Server::get(IAppManager::class)->isEnabledForUser('text') && class_exists(LoadEditor::class)) {
 			$this->eventDispatcher->dispatchTyped(new LoadEditor());
+		}
+
+		if (class_exists(LoadSidebar::class)) {
+			$this->eventDispatcher->dispatchTyped(new LoadSidebar());
+		}
+
+		if (\OCP\Server::get(IAppManager::class)->isEnabledForUser('viewer') && class_exists(LoadViewer::class)) {
+			$this->eventDispatcher->dispatchTyped(new LoadViewer());
 		}
 
 		$this->initialState->provideInitialState(
--- a/lib/Service/Note.php
+++ b/lib/Service/Note.php
@@ -105,6 +105,9 @@ class Note {
 		if (!in_array('readonly', $exclude)) {
 			$data['readonly'] = $this->getReadOnly();
 		}
+		$data['internalPath'] = $this->noteUtil->getPathForUser($this->file);
+		$data['shareTypes'] = $this->noteUtil->getShareTypes($this->file);
+		$data['isShared'] = (bool) count($data['shareTypes']);
 		$data['error'] = false;
 		$data['errorType'] = '';
 		if (!in_array('content', $exclude)) {
--- a/src/components/NoteItem.vue
+++ b/src/components/NoteItem.vue
@@ -26,9 +26,19 @@
 				fill-color="var(--color-text-lighter)"
 			/>
 		</template>
+		<template v-if="isShared" #indicator>
+			<ShareVariantIcon :size="16" fill-color="#0082c9" />
+		</template>
 		<template #actions>
 			<NcActionButton :icon="actionFavoriteIcon" @click="onToggleFavorite">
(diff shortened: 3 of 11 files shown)


composer.json
```diff
@@ -1,6 +1,6 @@
 {
 	"require-dev": {
-		"nextcloud/ocp": "dev-stable25@dev",
+		"nextcloud/ocp": "dev-master",
```
juliushaertl, line 3: Can you revert this specific change, currently we still cover multiple nextcloud versions and want to run the psalm checks against the ocp package towards the lowest supported.

lib/Controller/PageController.php
```diff
@@ -66,6 +68,14 @@ public function index() : TemplateResponse {
 			$this->eventDispatcher->dispatchTyped(new LoadEditor());
 		}
 
+		if (\OCP\Server::get(IAppManager::class)->isEnabledForUser('files') && class_exists(LoadSidebar::class)) {
```
juliushaertl, line 71: This should not be needed as the files app is force enabled and cannot be disabled, though maybe we just keep it if the server ever changes that

lib/Service/Note.php
```diff
@@ -105,6 +105,9 @@ public function getData(array $exclude = []) : array {
 		if (!in_array('readonly', $exclude)) {
 			$data['readonly'] = $this->getReadOnly();
 		}
+		$data['internalPath'] = $this->file->getInternalPath();
```
juliushaertl, line 108: I pushed a small fix for getting the proper relative path for the user to be used in the frontend.

lib/Service/Note.php
```diff
@@ -105,6 +105,9 @@ public function getData(array $exclude = []) : array {
...
+		$data['shareTypes'] = $this->noteUtil->getShareTypes($this->file);
```
juliushaertl, line 109: Should allow to achieve the same without triggering heavy loading of all the share entries from the db:
```suggestion
		$data['shareTypes'] = $this->file->isShared();
```

lib/Service/Note.php
```diff
@@ -105,6 +105,9 @@ public function getData(array $exclude = []) : array {
...
+		$data['isShared'] = (bool) count($data['shareTypes']);
```
juliushaertl, line 110: Should allow to achieve the same without triggering heavy loading of all the share entries from the db:

```suggestion
		$data['shareTypes'] = $this->file->isShared();
```

src/components/NoteItem.vue
```diff
@@ -168,6 +190,23 @@ export default {
...
 			}
 		},
+		onToggleSharing() {
+			if (window?.OCA?.Files?.Sidebar?.setActiveTab) {
+				emit('toggle-navigation', { open: false })
+				setTimeout(() => {
+					window.dispatchEvent(new Event('resize'))
+				}, 200)
+				window.OCA.Files.Sidebar.setActiveTab('sharing')
+				window.OCA.Files.Sidebar.open(this.note.internalPath.replace('files/', '/'))
```
juliushaertl, line 200: This might be tricky if the path returns files in the middle, but should not be necessary anymore with my pushed commit ;)
```suggestion
				window.OCA.Files.Sidebar.open(this.note.internalPath)
```

juliushaertl (changes requested):
Small comments, but otherwise this looks quite good 👍

src/components/NoteRich.vue
```diff
@@ -144,6 +147,38 @@ export default {
...
+			}
+
+			const autoResolve = setInterval(() => {
+				const el = document.querySelector('[data-cy="resolveServerVersion"]')
+
+				if (el) {
+					el.click()
+					clearInterval(autoResolve)
+				}
+			}, 200)
```
juliushaertl, line 179: This seems too hacky. I'd rather have a separate fix for this somehow in text.

composer.json
luka-nextcloud: Class `OCP\Share\Events\BeforeShareCreatedEvent` does not exist on `"nextcloud/ocp": "dev-stable25@dev",` . We are listening to this event [here](https://github.com/nextcloud/notes/pull/1146/files#diff-4eb824426a8cffb8b33b62b57744c69f32ea30ea755249a7a6170d9402885debR31)

lib/Service/Note.php
luka-nextcloud: I have tried `$this->file->isShared()` but it always returns false.

juliushaertl: Right, it would only return true for incoming shares (so for the receiving user). Let's keep that then as it is as the files app seems to do it similarly.

composer.json
juliushaertl: Right, it is only available since 28. 

Let me check if there is a fallback we can use for that then

juliushaertl: I pushed a fallback and added the event to the stub for the static analysis tests.

@luka-nextcloud I just noticed that my previously added commit was overwritten with yours and repushed that. Quick hint, if you use `git push --force-with-lease` instead of just a force push you will get notified about changes that have been made in the meantime. https://git-scm.com/docs/git-push#Documentation/git-push.txt---no-force-with-lease

juliushaertl (approved):

composer.json
luka-nextcloud: Ok, thanks for your tip.

Issue #1177, opened 2023-12-09
App Unable to Read Notes From Server After Update to 4.9


kn0wmad:
"Error from Nextcloud server: Reading notes from filesystem has failed. (Error)"

Attempted to remove and re-add folder path, attempted to uninstall and re-install app, neither fixed issue

## Steps to reproduce
1. Update to latest Notes
2. Visit Notes web app in NC
3. See error

### Expected behaviour
Notes should be read from FS

### Actual behaviour
Notes unable to read existing path

## Server
Please complete the following information.

- Notes app version: 4.9
- Nextcloud version: 25.0.5
- OS: Alpine
- Web server: nginx
- PHP version: 8.1
- Database: Postgres

**Nextcloud configuration:**

<details>
```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder
```
</details>

## Client
Please complete the following information.

- Browser (incl. version): FF 120
 - OS: Linux 6.1

## Log files
Add relevant parts of your `nextcloud.log` and/or your browser's JavaScript console here.

juliushaertl:
> Add relevant parts of your nextcloud.log and/or your browser's JavaScript console here.

Can you check the logs?

GuestJim:
I'm getting the same error message when I try using Notes, and here's what it shows in the logs for each attempt:

---
[notes] Error: Error: Undefined constant OCP\Share\IShare::TYPE_SCIENCEMESH at <<closure>>

 0. /var/www/html/custom_apps/notes/lib/Service/Note.php line 109
    OCA\Notes\Service\NoteUtil->getShareTypes(OC\Files\Node\File {})
 1. /var/www/html/custom_apps/notes/lib/Controller/Helper.php line 62
    OCA\Notes\Service\Note->getData(["etag","content"])
 2. /var/www/html/custom_apps/notes/lib/Controller/Helper.php line 114
    OCA\Notes\Controller\Helper->getNoteData(OCA\Notes\Service\Note {}, ["etag","content"], OCA\Notes\Db\Meta {id: 25089})
 3. <<closure>>
    OCA\Notes\Controller\Helper->OCA\Notes\Controller\{closure}("*** sensitive parameters replaced ***")
 4. /var/www/html/custom_apps/notes/lib/Controller/Helper.php line 113
    array_map(Closure {}, ["*** sensitive  ... "])
 5. /var/www/html/custom_apps/notes/lib/Controller/NotesController.php line 71
[...]

PR #1182, opened 2023-12-15
fix: Avoid using constant that is not available on 25

juliushaertl:
Fix https://github.com/nextcloud/notes/issues/1177 for stable25


datenangebot (approved):
Looks reasonable.

Btw: I guess "FIXME" is also a keyword that gets tracked?
Otherwise I would "TODO" recommend as this might be tracked in most IDEs...

juliushaertl:
FIXME is also collected by phpstorm for example (together with TODO)

PR #1191, opened 2023-12-25
Add bidi support

ahangarha:
This PR is a part of a series of PRs to the NextCloud ecosystem to add bidirectional text support.

In terms of the overall layout, this PR doesn't make any changes in the UI in LTR language (and in the current state of the NextCloud server for RTL language). But it brings flexibility to the UI so that in the RTL layout, it makes automatic adjustments.

Since the editor is a third-party package, adding bidi support to it will take longer time. We might even consider other options as codemirror failed in adding bidi support for a long time. But for the Preview, I added `markdown-it-bidi` which is working fine now.

### Screenshot (in RTL layout)
image

### DIFF
--- a/css/notes.css
+++ b/css/notes.css
@@ -28,20 +28,28 @@ li:focus  .app-navigation-entry__utils .action-item {
 
 /* Welcome / Help */
 .feature {
-	background-position: left center;
+	background-position-y: center;
 	width: 100%;
 	min-height: 32px;
 	line-height: 32px;
-	padding-left: 32px;
+	padding-inline-start: 32px;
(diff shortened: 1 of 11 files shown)


luka-nextcloud (approved):

juliushaertl:
Thanks so much for this contribution @ahangarha

ahangarha:
Thanks for your review and merge.

Is there any particular reason to stick to a markdown editor based on codemirror? I am unsure how long it would take for codemirror to add bidi support.

If not, I think there are good, viable options to consider.

PR #1260, opened 2024-03-27
fix: Avoid conflicts on notes folder creation

juliushaertl:


mejo- (approved):

Waltibaba:
This breaks the Notes app when the parent is a shared folder.
If selected folder is shared, create new folder. New folder is of course also shared (since the parent is shared), so it may not be used -> create new folder.
On a fast system that syncs often, this just creates hundreds of new folders in minutes.
This caused #1266

PR #1263, opened 2024-03-28
fix: Shared folder check

provokateurin:
Fixes `Call to a member function isShared() on null in file '/usr/src/nextcloud/apps/notes/lib/Service/NoteUtil.php' line 171` introduced by https://github.com/nextcloud/notes/pull/1260.

This is triggered when the user has no notes folder yet and opens the dashboard (at least that was how I triggered it).

### DIFF
--- a/lib/Service/NoteUtil.php
+++ b/lib/Service/NoteUtil.php
@@ -168,14 +168,15 @@ class NoteUtil {
 			$folder = $this->root->newFolder($path);
 		}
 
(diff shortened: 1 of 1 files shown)


provokateurin:
@juliushaertl if possible this should be released as 4.9.4 because 4.9.3 is currently broken.

juliushaertl (approved):

juliushaertl:
Sure, I'll spin another release with that. Thanks for the fix 🙏

Issue #1259, opened 2024-03-27
"Uncaught (in promise) TypeError: t is undefined" after updating to nextcloud 28.0.3 and latest notes


HolgerHees:
Environment:
- nextcloud 28.0.3
- notes (latest version)
- firefox 124

After updating to the latest version I get the trace at the end (TRACE 1).

After that, the ui is not useable anymore. e.g. if I click on new note, nothing happens with a lot more exceptions.

For me it looks like a timing problem, because it does not happen always.

 (TRACE 1)
```
09:12:53.929 Uncaught (in promise) TypeError: t is undefined
    render editor.js:77
    createEditor editor.js:234
    loadTextEditor NoteRich.vue:89
    fetchData NoteRich.vue:79
    created NoteRich.vue:56
    VueJS 28
editor.js:77:2
    fetchData NoteRich.vue:79
    AsyncFunctionThrow self-hosted:857
    (Async: async)
    created NoteRich.vue:56
    VueJS 69
    init vue-router.esm.js:3005
    forEach self-hosted:203
    init vue-router.esm.js:3004
    updateRoute vue-router.esm.js:2414
    transitionTo vue-router.esm.js:2263
    confirmTransition vue-router.esm.js:2402
    r vue-router.esm.js:2084
    r vue-router.esm.js:2091
    Dt vue-router.esm.js:2095
    confirmTransition vue-router.esm.js:2396
    r vue-router.esm.js:2084
    r vue-router.esm.js:2088
    p vue-router.esm.js:2384
    h vue-router.esm.js:2162
    p vue-router.esm.js:2362
    r vue-router.esm.js:2087
    r vue-router.esm.js:2091
    r vue-router.esm.js:2091
    Dt vue-router.esm.js:2095
    confirmTransition vue-router.esm.js:2392
    transitionTo vue-router.esm.js:2260
    push vue-router.esm.js:2606
    push vue-router.esm.js:3036
    push vue-router.esm.js:3035
    routeToNote App.vue:215
    routeDefault App.vue:194
    loadNotes App.vue:137
    (Async: promise callback)
    loadNotes App.vue:130
    created App.vue:119
    VueJS 17
    <anonym> main.js:12
    <anonym> main.js:16
    <anonym> main.js:16
```

HolgerHees:
here comes a small update.

If I open any other app (like the files app) and open afterwards the notes app, the bug is 100% reproducible

If I reload the notes app by pressing the browser reload button the bug does not happen

[...]

PR #1264, opened 2024-03-29
Fix "TypeError: t is undefined"

HolgerHees:
vue "created" does not guaranteed that all refs are filled. Sometime $refs.editor is undefined. Also an extra tick like in line 84 does not help. 

created() is great for calling APIs, while mounted() is great for doing anything after the DOM elements have completely loaded.

using "mounted" instead of "created" fixes this issues

the bug itself is documented here

https://github.com/nextcloud/notes/issues/1259

### DIFF
--- a/src/components/NoteRich.vue
+++ b/src/components/NoteRich.vue
@@ -52,7 +52,7 @@ export default {
 		},
 	},
 
-	created() {
+	mounted() {
 		this.fetchData()
 		subscribe('files:file:updated', this.fileUpdated)
 		subscribe('files_versions:restore:requested', this.onFileRestoreRequested)


HolgerHees:
@juliushaertl is there anything missing or something I can do?

jancborchardt:
@korelstar could you verify this? :) Also @mejo- @max-nextcloud maybe one of you can check.

juliushaertl (approved):
Thanks a lot, makes sense 👍

juliushaertl:
CI failures are unrelated, will check separately

mchiappinam:
Is that issue with Notes + Nextcloud back on version 30? I was on version 28 today, upgraded to 29 and at that point the notes app was working still. But after upgrading to v30, I'm seeing this:
NoteRich.vue:89 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'createEditor')
image
Note: I don't think the first error is related as it says bootstrap.

max-nextcloud:
@mchiappinam That's most likely a different issue. Your exception says `createEditor` is called on `undefined`. In this case `t` was `undefined` but `createEditor` is not called on `t` - so the two exceptions are not related. Plus this fix has been merged a while back. So it's probably included in Nextcloud 30 already.

Please open a separate issue.

Issue #1266, opened 2024-03-31
Cannot use shared folder for notes anymore. 


SimeonEhrig:
I share the `Notes` between two accounts. The first account is my main account. The second account is a dedicated account for my smartphone. Up to version 4.9.3, everything works fine. I could create and edit markdown files on my PC and sync it via Nextcloud client to the main account, use the web interface of my main account for editing notices and read and write notices on my smartphone with the [nextcloud notes app](https://f-droid.org/en/packages/it.niedermann.owncloud.notes/) syncing over the dedicated android account.   

Since version 4.9.4. The android app and the notice app of the android account is not working anymore. The notice app on the main account is still working.

I think, this is related to this PR: https://github.com/nextcloud/notes/pull/1263

## Steps to reproduce
1. Create two accounts on the same nextcloud instance
2. share the `Notes` folder of the account A with account B
3. try to create a note in the web interface of account B and check if it is available on account A

### Expected behaviour

The notices should be shared and the web interface and the android app should work with account A and B in the same way.

### Actual behaviour
Account B creates an new notes folder (`Notes (2)`) and stores the notes there.

## Server
Please complete the following information.

- Notes app version: 4.9.4
- Nextcloud version: 28.0.4
- OS: Ubuntu 23.10 server
- Web server: Apache (builtin in nextcloud docker image: https://github.com/nextcloud/docker)
- PHP version: 8.2.17
- Database: mariadb 10.6.17

**Nextcloud configuration:**

```php
{
    "system": {
        "instanceid": "***REMOVED SENSITIVE VALUE***",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "nextcloud.domain.com"
        ],
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
        "overwrite.cli.url": "https:\/\/nextcloud.domain.com",
        "overwriteprotocol": "https",
        "dbtype": "mysql",
        "version": "28.0.4.1",
        "dbname": "***REMOVED SENSITIVE VALUE***",
        "dbhost": "***REMOVED SENSITIVE VALUE***",
        "dbport": "",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "logtimezone": "UTC",
        "installed": true,
        "maintenance": false,
[...]

PR #1296, opened 2024-05-11
allow using shared notes folder

SimeonEhrig:
Enables the use of a shared folder as a notes folder. Reverts PR #1260 as it is not clear what the problem was and using a shared notes folder is a feature that many users use: #1266

### DIFF
--- a/lib/Service/NoteUtil.php
+++ b/lib/Service/NoteUtil.php
@@ -172,11 +172,6 @@ class NoteUtil {
 			throw new NotesFolderException($path.' is not a folder');
 		}
 
-		if ($folder->isShared()) {
-			$folderName = $this->root->getNonExistingName($path);
-			$folder = $this->root->newFolder($folderName);
-		}
-
 		return $folder;
 	}
 


korelstar (approved):

LMRW:
Please I would love to see this fix merged in!

LMRW:
Can I help get this merged? Let me know if I can help.

This would fix a critical issue for me which i've spent a lot of time on

LMRW:
What can we do to fix this to be merged in please?

Its a fix for a show stopping bug that breaks a critical feature

Its also old tested code used before which was removed and would be put back, so its battle tested

Please there seems to be no good reason to not merge this in now

SimeonEhrig:
Any progress here?

juliushaertl:
Thanks for the pull request, unfortunately this reintroduces an issue where users without a notes folder could end up with their notes in an incoming share by accident. I've pushed a different approach to the fix to https://github.com/nextcloud/notes/pull/1320 that I'll try to finish up asap and do a new release then.

SimeonEhrig:
close because PR #1320 solved the problem.

PR #1320, opened 2024-07-11
fix/shared notes

juliushaertl:
The current backend logic to get or create a notes folder is not ideal as we cannot easily determine if a folder exists because a user has created and set is as notes folder before or if it was manually picked as a notes folder.

To avoid issues with the wrong incoming shared folder being taken as notes folder automatically this PR refactors the logic so that we now no longer store the notes folder in the config with the default value, until it was actually created. That way we know as soon as the config value is set, we no longer need to check for incoming shares on the folder. 

Fix https://github.com/nextcloud/notes/issues/1266

### DIFF
--- a/lib/AppInfo/BeforeShareCreatedListener.php
+++ b/lib/AppInfo/BeforeShareCreatedListener.php
@@ -49,8 +49,7 @@ class BeforeShareCreatedListener implements IEventListener {
 			$receiver = $share->getSharedWith();
 			$receiverPath = $this->noteUtil->getRoot()->getUserFolder($receiver)->getPath();
 			$receiverNotesInternalPath = $this->settings->get($receiver, 'notesPath');
-			$receiverNotesPath = $receiverPath . '/' . $receiverNotesInternalPath;
-			$this->noteUtil->getOrCreateFolder($receiverNotesPath);
+			$this->noteUtil->getOrCreateNotesFolder($receiver);
 
 			if ($itemType !== 'file' || strpos($fileSourcePath, $ownerNotesPath) !== 0) {
 				return;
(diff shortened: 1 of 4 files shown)


luka-nextcloud (commented):
Code change looks ok. Except the linting issue at https://github.com/nextcloud/notes/pull/1320/files#diff-7f7d9bb77038513c05d8a832f52c8b4ef5a00af9373c6b2fdaaf24b5b7d907d3R33

lib/AppInfo/BeforeShareCreatedListener.php
```diff
@@ -50,7 +50,7 @@ public function overwriteShareTarget(IShare $share): void {
 			$receiverPath = $this->noteUtil->getRoot()->getUserFolder($receiver)->getPath();
 			$receiverNotesInternalPath = $this->settings->get($receiver, 'notesPath');
 			$receiverNotesPath = $receiverPath . '/' . $receiverNotesInternalPath;
-			$this->noteUtil->getOrCreateFolder($receiverNotesPath);
+			$this->noteUtil->getOrCreateNotesFolder($receiverNotesPath, true);
```
luka-nextcloud, line 53:
```suggestion
			$this->noteUtil->getOrCreateNotesFolder($receiver, true);
```

juliushaertl: Good catch 👍

luka-nextcloud (approved):

SimeonEhrig:
@juliushaertl It's working, but I need a small workaround and I'm not sure, if this expected. If I share my folder from my main account to my android account, the shared folder is also named `Notes` in the android account. In this case, the note app tries to create a new folder ( `Notes (2)`). If I rename the shared folder in the android account to `Notes_Simeon` (the main account is not affected) and select it as notes folder in the notes app settings, everything is working. So, is this expected?

Issue #1325, opened 2024-07-15
Add SPDX headers


juliushaertl:
We should add SPDX headers and license information to be able to verify reuse compliance.

Examples can be found in other apps like text.

Useful script: https://github.com/nextcloud/github_helper/tree/master/spdx-convertor

PR #1360, opened 2024-09-10
feat: add SPDX headers

luka-nextcloud:
Resolves: #1325

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,9 @@
+<!--
+  - SPDX-FileCopyrightText: 2019-2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
 # Notes
+[![REUSE status](https://api.reuse.software/badge/github.com/nextcloud/notes)](https://api.reuse.software/info/github.com/nextcloud/notes)
 
 <!-- The following paragraph should be kept synchronized with the description in appinfo/info.xml -->
 The Notes app is a distraction free notes taking app for [Nextcloud](https://www.nextcloud.com/). It provides categories for better organization and supports formatting using [Markdown](https://en.wikipedia.org/wiki/Markdown) syntax. Notes are saved as files in your Nextcloud, so you can view and edit them with every Nextcloud client. Furthermore, a separate [REST API](https://github.com/nextcloud/notes/blob/master/docs/api/README.md) allows for an easy integration into apps ([Android](https://github.com/nextcloud/notes-android), [iOS](https://github.com/nextcloud/notes-ios), and 3rd-party apps like the [console](https://git.danielmoch.com/nncli/about) and [many more](https://github.com/nextcloud/notes/wiki#3rd-party-clients) which allow convenient access to your Nextcloud notes). Further features include marking notes as favorites.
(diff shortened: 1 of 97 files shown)


juliushaertl:
We also need:
- [ ] Add a github action to check this https://github.com/nextcloud/.github/blob/master/workflow-templates/reuse.yml
- [ ] Make sure that the check passes
- [ ] Add an AUTHORS file (see https://github.com/nextcloud/text/pull/5738)
- [ ] Have a reuse config file for cases where we cannot define the license in the headers (see https://github.com/nextcloud/whiteboard/pull/69/files)

appinfo/routes.php
```diff
@@ -1,5 +1,10 @@
 <?php
 
+/**
+ * SPDX-FileCopyrightText: 2013 Nextcloud GmbH and Nextcloud contributors
```
juliushaertl, line 4: Need to double check if that is OK, or if we need to have a different copyright header since 2013 was before Nextcloud

luka-nextcloud: There is copyright from first commit: https://github.com/nextcloud/notes/commit/1c4ae20dad288b5691e87a1fabb5e944071ca782#diff-6301937af1cc1575e9e18ecd27c4a51f0d3cf9ce12e59a14709f626dfa3ef952R2-R6

juliushaertl: Maybe @AndyScherzinger can easily tell if we can attribute that as `Nextcloud GmbH and Nextcloud contributors`or if we need to keep Bernhard explicitly in there

AndyScherzinger: I'd say keep him explicitly there. Given code existing before Nextcloud has been founded, Nextcloud cannot be a Copyright holder in any way.
So for anything oder than 2016, best to keep the original copyright holder Infos in additionally.

appinfo/info.xml
```diff
@@ -1,4 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!--
+  - SPDX-FileCopyrightText: 2013-2024 Nextcloud GmbH and Nextcloud contributors
```
AndyScherzinger, line 3: Should be changed to 
SPDX-FileCopyrightText: 2016-2024 Nextcloud GmbH and Nextcloud contributors
SPDX-FileCopyrightText: 2013 Bernhard Posselt <nukeawhale@gmail.com>

README.md
```diff
@@ -1,3 +1,7 @@
+<!--
+  - SPDX-FileCopyrightText: 2019-2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
 # Notes
```
AndyScherzinger, line 6: Also please add the status badge here via:
```
[![REUSE status](https://api.reuse.software/badge/github.com/nextcloud/notes)](https://api.reuse.software/info/github.com/nextcloud/notes)
```

AndyScherzinger:
Changes look good to me reuse/spdx-wise. Just 2 minor comments else 👍

AndyScherzinger (approved):

juliushaertl (approved):

PR #1387, opened 2024-10-10
fix: Adjust note item size to be one line

### LINKED ISSUE #1383: Too much wasted space in list of notes
When using the webGUI for Notes, the left side which lists the notes wastes so much vertical space on presentation. Take a look:

image

The area above and below "Workout notes" and "Groceries" is effectively in total triple the actual space used by the text itself.

This leads to very few items visible on-screen, despite using higher resolution monitors (such as 2560x1440 in my case).

This really is an obnoxious, wasteful, and inefficient design. And I for one would really want to see a lot less of a gap above and below the name of each note in the left list. I'm not saying the gap should be eliminated, and I'm not even sure what I might consider the ideal amount. But maybe half to start?

Also, this is with the latest version as of writing.

Either way, as it stands this really is pretty yucky.

juliushaertl:
Otherwise lines are too large with the new more compact design

| Before | After |
|---|---|
| Screenshot 2024-10-10 at 16 52 23 | Screenshot 2024-10-10 at 16 52 32 |

Fixes #1383

### DIFF
--- a/src/components/NoteItem.vue
+++ b/src/components/NoteItem.vue
@@ -8,6 +8,7 @@
 		:name="title"
 		:active="isSelected"
 		:to="{ name: 'note', params: { noteId: note.id.toString() } }"
+		one-line
 		@update:menuOpen="onMenuChange"
 		@click="onNoteSelected(note.id)"
 	>


juliushaertl:
Not intentionally, I pushed https://github.com/nextcloud/notes/pull/1389 to standardize the ci checks a bit and make them pass again

BloodyIron:
Thanks! :D

BloodyIron:
Would we be able to get a v4.11.1 released with this and other commits to the repo? Or something like that? Really hoping this can hit my app update options sooner rather than later :)

Also, oooo to the corner changes! I didn't notice that before :)

PR #1410, opened 2024-10-31
Migrate notes search to NcTextField with clear button

korelstar:
The search input field for searching notes did not have a button for clearing the input value. However, this is standard in many input fields used in Nextcloud. Therefore, this PR migrates the HTML input field into a `NcTextField` which automatically has a button for clearing the input value.

Before:
Screenshot

After:
Screenshot


JuliaKirschenheuter (approved):
Thanks for taking over!

PR #1423, opened 2024-11-23
Micro text change: Remove reference to https://github.com/djmoch/nncli

per-oestergaard:
As the repo was archived Jun 18, 2024


juliusknorr:
I'd say this change makes sense. We should not advertise unmaintained projects to prominently. We can still have a reference with a hint that it is no longer maintained in https://github.com/nextcloud/notes/wiki

per-oestergaard:
> @per-oestergaard Can you please update the commit messages to have the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format so that this CI test passes https://github.com/nextcloud/notes/actions/runs/12607582648/job/35139116282?pr=1423. For example `Remove reference to nncli` -> `docs: Remove reference to nncli`

Yes, no problem. But I am not sure how I do that as I have never tried that git variant before. Do I have to change the text of both? Or can I create an empty commit with the right format? Or can it be fixed when merging?

juliusknorr:
For a single commit, you can use `git commit --amend` to add to your existing commit and change the commit message accordingly. However since you have two the easiest would be to squash them first and then update the commit message using an [interactive rebase](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History). Do you want to give that a try? Otherwise we can also do that while merging.

per-oestergaard:
> For a single commit, you can use `git commit --amend` to add to your existing commit and change the commit message accordingly. However since you have two the easiest would be to squash them first and then update the commit message using an [interactive rebase](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History). Do you want to give that a try? Otherwise we can also do that while merging.

Better now?

juliusknorr:
Looks good, thanks. Lets merge this, CI failures are unrelated

PR #1460, opened 2025-01-09
Fix note caption colour for custom themes

looowizz:
Custom themes may use colours that do not contrast well with the background for their primary element colour, e.g. [my theme](https://codeberg.org/Loowiz/nextcloud-possums-theme)

Removing the line here means the default text colour is used, which I think should be in use here anyway?

### DIFF
--- a/src/components/NotesCaption.vue
+++ b/src/components/NotesCaption.vue
@@ -28,7 +28,6 @@ export default {
 <style scoped>
 	.name {
 		font-weight: bold;
-		color: var(--color-primary-element);
 		font-size: var(--default-font-size);
 		line-height: var(--default-clickable-area);
 		white-space: nowrap;


looowizz:
Sorry, probably should've included this initially.

I think the change gives users more freedom to choose a wider range colours for their primary colour, and I think it just makes more sense as the main text colour is also used in all other Nextcloud header text.

## Custom theme
### Before
customThemeBefore
### After
customThemeAfter

## Default theme
### Before
normalThemeBefore
### After
normalThemeAfter

looowizz:
> Thanks, looks good.
> 
> Can you pleae amend your commit message to follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format, so that the [CI test](https://github.com/nextcloud/notes/actions/runs/12699138243/job/36157057971?pr=1460) passes?

@enjeck Done! :)

PR #1468, opened 2025-01-24
add notes path to capabilities

tobiasKaminsky:


lib/AppInfo/Capabilities.php
```diff
@@ -9,21 +9,23 @@
...
+								private NoteUtil $noteUtil,
+								private IUserSession $userSession) {
 	}
 
 	public function getCapabilities(): array {
 		return [
 			Application::APP_ID => [
 				'api_version' => Application::$API_VERSIONS,
 				'version' => $this->appManager->getAppVersion(Application::APP_ID),
+				'notes_path' => $this->noteUtil->getOrCreateNotesFolder($this->userSession->getUser()->getUID())->getInternalPath()
```
szaimen, line 28: Maybe this?

```suggestion
				'notes_path' => $this->noteUtil->getOrCreateNotesFolder($this->userSession->getUser()?->getUID())->getInternalPath()
```

tobiasKaminsky: This also gives /files/Notizen" , but not "/Notizen"

tobiasKaminsky:
@juliusknorr thanks! Tested & works!

tobiasKaminsky:
@juliusknorr or @enjeck can you help me with psalm?

```
ERROR: RiskyTruthyFalsyComparison
at /home/tobi/github/nextcloud/notes/lib/AppInfo/Capabilities.php:29:21
Operand of type null|string contains type string, which can be falsy and truthy. This can cause possibly unexpected behavior. Use strict comparison instead. (see https://psalm.dev/356)
                                'notes_path' => $this->userId ? $this->noteUtil->getNotesFolderUserPath($this->userId) : null,
```

Not sure what to do.

tobiasKaminsky:
@juliusknorr I am sorry, but I give up on this.
I am not familar with it, and having two lines of code changes, killing the entire CI… :shrug:

lib/Service/NoteUtil.php
```diff
@@ -183,6 +183,13 @@ public function getOrCreateFolder(string $path, bool $create = true) : Folder {
 		return $folder;
 	}
 
+	public function getNotesFolderUserPath(string $userId): ?string {
+		/** @psalm-suppress MissingDependency */
+		$userFolder = $this->getRoot()->getUserFolder($userId);
+		$nodesFolder = $this->getOrCreateNotesFolder($userId, false);
```
blizzz, line 189: This is probably on users where the folder does not exists yet (e.g. new users on first login). 
Possible solutions: either the `create` parameter has to be true, or the `NotesFolderException` has to be caught.

blizzz: An integration test would have caught this btw.

PR #1530, opened 2025-04-26
fix: Use info instead of warning to avoid filling logs

enjeck:
fixes https://github.com/nextcloud/notes/issues/1526

### DIFF
--- a/lib/Service/NoteUtil.php
+++ b/lib/Service/NoteUtil.php
@@ -189,7 +189,7 @@ class NoteUtil {
 		try {
 			$nodesFolder = $this->getOrCreateNotesFolder($userId, false);
 		} catch (NotesFolderException $e) {
-			$this->util->logger->info("Failed to get notes folder for user $userId: " . $e->getMessage());
+			$this->util->logger->debug("Failed to get notes folder for user $userId: " . $e->getMessage());
 			return null;
 		}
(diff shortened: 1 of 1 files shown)


lib/Service/NoteUtil.php
```diff
@@ -189,7 +189,7 @@ public function getNotesFolderUserPath(string $userId): ?string {
 		try {
 			$nodesFolder = $this->getOrCreateNotesFolder($userId, false);
 		} catch (NotesFolderException $e) {
-			$this->util->logger->warning("Failed to get notes folder for user $userId: " . $e->getMessage());
+			$this->util->logger->info("Failed to get notes folder for user $userId: " . $e->getMessage());
```
joshtrichards, line 192:
```suggestion
			$this->util->logger->info("Failed to get notes folder for user $userId: " . $e->getMessage());
```

`info` is acceptable; `debug` better IMO since this isn't necessarily always a sign of any sort of problem

joshtrichards (approved):

JuliaKirschenheuter (approved):

lib/Service/NoteUtil.php
juliusknorr: Agreed in this case as we do not create it here automatically

Issue #1245, opened 2024-03-06
Hide hidden directory 


rgischig:
**Is your feature request related to a problem? Please describe.**
I'm using a note taking app which created a .trash file inside the main folder to keep the old or deleted files. Now in the Notes direcotry viewer, I get the .trash direcitry, which has a lot of files in it, as my first option! I need to scroll down almost to the bottom to see my real content.

**Describe the solution you'd like**
option to hide hidden directories and files

**Describe alternatives you've considered**
sort-order to have hidden files at the end of the list

**Additional context**
![Screenshot 2024-03-06 082252](https://github.com/nextcloud/notes/assets/21161212/f8576c95-db33-4304-9ca0-0b4a3211c623)

joshtrichards:
For the record, what note taking app?

rgischig:
The Notes app of Nextcloud:
https://github.com/nextcloud/notes

My notes in the example where generated by Obsidian, which created these .trash folders as backup inside the vault, so you would have access on any client to the deleted files...

k3dar:
same problem with Syncthing .stversions dir, all files in it is show in Notes app (in stock Web App and also in Android App)

is posible change in Syncthing DIR setting, path for versioning outside synced dir, although it's not very comfortable,
but .stfolder must be inside synced dir, and Notes app show syncthing-folder-xxyyzz.txt as Note...

can be please added option to ignore/hide hidden/dot directory?

PR #1541, opened 2025-05-09
feat: setting to show/hide hidden files

jrvidal:
This is a fix for #1245. It adds a new `showHidden` setting to control whether hidden files/folders are displayed to the user or not.

* "Hidden" means "starts with a dot", a la UNIX. I'm not sure how much we care about a hypothetical notes folder living within a Windows FS that contains windows-like hidden files. (I supposed the answer is "not much").

* The setting is only respected in the main web interface, and ignored for API methods. This is a quick-n-dirty compromise between letting clients handle hidden folders/files as they see fit on their own; vs. exposing this setting as public and store it _per client_ (which goes beyond my knowledge of NC internals).

* The setting also affects other entrypoints that transitively depend on `NotesService#getAll()`, such as the search provider and the top-level notes. This seems appropriate to me, given its main usecase is probably metadata files generated by 3rd party apps.

* I've set the default to `true` so it is "backwards compatible". I'm not sure how acceptable would be to switch it to `false`.

* I haven't been able to get a proper setup to run tests locally, but I do want to add some tests.

### DIFF
--- a/lib/Service/NotesService.php
+++ b/lib/Service/NotesService.php
@@ -23,11 +23,14 @@ class NotesService {
 	) {
 	}
 
-	public function getAll(string $userId, bool $autoCreateNotesFolder = false) : array {
+	public function getAll(string $userId, bool $autoCreateNotesFolder = false, ?bool $showHidden = null) : array {
 		$customExtension = $this->getCustomExtension($userId);
 		try {
 			$notesFolder = $this->getNotesFolder($userId, $autoCreateNotesFolder);
-			$data = self::gatherNoteFiles($customExtension, $notesFolder);
+			if ($showHidden === null) {
+				$showHidden = $this->settings->get($userId, 'showHidden');
+			}
+			$data = self::gatherNoteFiles($customExtension, $notesFolder, $showHidden);
 			$fileIds = array_keys($data['files']);
 			// pre-load tags for all notes (performance improvement)
 			$this->noteUtil->getTagService()->loadTags($fileIds);
@@ -59,7 +62,8 @@ class NotesService {
 		$customExtension = $this->getCustomExtension($userId);
 		try {
 			$notesFolder = $this->getNotesFolder($userId, false);
-			$data = self::gatherNoteFiles($customExtension, $notesFolder);
+			$showHidden = $this->settings->get($userId, 'showHidden');
+			$data = self::gatherNoteFiles($customExtension, $notesFolder, $showHidden);
 			return count($data['files']);
 		} catch (NotesFolderException $e) {
 			return 0;
@@ -241,6 +245,7 @@ class NotesService {
 	private static function gatherNoteFiles(
 		string $customExtension,
 		Folder $folder,
+		bool $showHidden,
 		string $categoryPrefix = '',
 	) : array {
 		$data = [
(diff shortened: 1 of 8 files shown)


src/components/AppSettings.vue
```diff
@@ -195,6 +207,12 @@ export default {
 		setSettingsOpen(newValue) {
 			this.settingsOpen = newValue
 			this.$emit('update:open', newValue)
+
+			if (this.settingsOpen) {
+				this.$data.initialShowHidden = Boolean(store.state.app.settings.showHidden)
+			} else if (this.$data.initialShowHidden !== store.state.app.settings.showHidden) {
+				this.$emit('reload')
+			}
```
jrvidal, line 215: I'm getting fancy here and trying to avoid full in-place reloads until they're absolutely needed, instead of using `onChangeSettingsReload()`. I'm not sure it's a good fit for NC's general UX approach.

lib/Service/NotesService.php
```diff
@@ -30,11 +30,14 @@ public function __construct(
...
 	}
 
-	public function getAll(string $userId, bool $autoCreateNotesFolder = false) : array {
+	public function getAll(string $userId, bool $autoCreateNotesFolder = false, ?bool $showHidden = null) : array {
 		$customExtension = $this->getCustomExtension($userId);
 		try {
 			$notesFolder = $this->getNotesFolder($userId, $autoCreateNotesFolder);
-			$data = self::gatherNoteFiles($customExtension, $notesFolder);
+			if ($showHidden === null) {
+				$showHidden = $this->settings->get($userId, 'showHidden');
```
jrvidal, line 38: We now incur in an extra `SettingsService->get()` call (here and in `countNotes()`), but I suspect/expect (without deep knowledge of NC's internals) that settings retrieval might be backed by some cache?

enjeck (commented):
Looks good, just nitpicks that you can ignore

src/components/AppSettings.vue
```diff
@@ -22,17 +22,26 @@
 				{{ t('notes', 'Organize your notes in categories.') }}
 			</div>
 		</NcAppSettingsSection>
-		<NcAppSettingsSection id="notes-path-section" :name="t('notes', 'Notes path')">
+		<NcAppSettingsSection id="notes-path-section" :name="t('notes', 'Notes folder')">
```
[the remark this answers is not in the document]
jrvidal: :+1: I changed it b/c it is now the section header for 2 controls instead of just the path input. But no hard opinion.

jrvidal:
@juliusknorr 


> For the mobile apps we could extend the API within notes to also pass the value along in
> 
> https://github.com/nextcloud/notes/blob/e19b79bafd8446852f8f7da4eee9bd844b8a06d9/lib/Controller/NotesApiController.php#L70
> As long as the mobile apps for notes are using the dedicated endpoints and not WebDAV I'd say they can also respect the setting.

Sorry I'm a bit confused by this comment. Are you saying we should _not_ hardcode `$showHidden = true` for the API endpoints, and read from settings instead? What do you mean by the "API within notes"?

Re: mobile apps using WebDAV: I assume you mean that we have no control over apps that use WebDAV, correct? i.e. that there's no way to propagate this setting for them (I assume the WebDAV handling is part of `files` itself).

juliusknorr:
> Sorry I'm a bit confused by this comment. Are you saying we should not hardcode $showHidden = true for the API endpoints, and read from settings instead? What do you mean by the "API within notes"?

Yes, sorry if that was not clear enough. I'd read the setting there instead of the hardcoded one as well.

> Re: mobile apps using WebDAV: I assume you mean that we have no control over apps that use WebDAV, correct? i.e. that there's no way to propagate this setting for them (I assume the WebDAV handling is part of files itself).

Yes, that is in the responsibility of webdav / Nextcloud server then to handle.

jrvidal:
:wave: I've merged `main` into this and fixed all the evident conflicts, but I haven't fully re-reviewed to catch errors due to stale assumptions.

enjeck:
@jrvidal I fixed some merge conflicts. Also, after https://github.com/nextcloud/notes/pull/1984 was merged, some of the functionality was done too

enjeck (approved):
thank you!

PR #1543, opened 2025-05-13
Fix slow capabilities

enjeck:
- Instead of calling `getOrCreateNotesFolder` which can be expensive, we just use the value in the cache `settingsService->get($userId, 'notesPath')` since we assume the result is the same
- Ensure that when `getDefaultNotesPath()` is called, the value is saved in the user settings such that we need not call it again since the function does file operations

### DIFF
--- a/lib/Service/SettingsService.php
+++ b/lib/Service/SettingsService.php
@@ -152,22 +152,18 @@ class SettingsService {
 		$settings = $this->getSettingsFromDB($uid);
 		// use default for empty settings
 		$toBeSaved = false;
-		$toSaveNotesPath = false;
 		foreach ($this->attrs as $name => $attr) {
 			if (!property_exists($settings, $name)) {
 				$defaultValue = $attr['default'];
 				if (is_callable($defaultValue)) {
 					$settings->{$name} = $defaultValue($uid);
 					$toBeSaved = $saveInitial;
-					if ($name === 'notesPath') {
-						$toSaveNotesPath = true;
-					}
 				} else {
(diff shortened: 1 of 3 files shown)


lib/Service/SettingsService.php
```diff
@@ -149,18 +152,22 @@ public function getAll(string $uid, $saveInitial = false) : \stdClass {
 		$settings = $this->getSettingsFromDB($uid);
 		// use default for empty settings
 		$toBeSaved = false;
+		$toSaveNotesPath = false;
```
juliusknorr, line 155: I'm not a fan of the special handling here, can't we just call this method with saveInitial set to true in case of the capabilities?

lib/Service/SettingsService.php
```diff
@@ -112,6 +112,9 @@ public function set(string $uid, array $settings, bool $writeDefaults = false) :
 			if ($value !== null && array_key_exists($name, $this->attrs)) {
 				$settings[$name] = $value = $this->attrs[$name]['validate']($value);
 			}
+			if ($name === 'notesPath' && $value !== null) {
```
juliusknorr, line 115: Do we need to limit this to the notesPath or could we just skip getting the default if the value is not null?

juliusknorr (approved):

nigelharpur:
Hi and thanks for all the work going on to fix stuff - I am currently on stable 31.0.5 - is there now an ETA for the combined issues to have a fix implemented? (my own issue is the "Failed to get notes folder for user..." log entry thing.)

PR #1556, opened 2025-05-23
build(release): 4.12.1

enjeck:


juliusknorr (changes requested):
Why not 4.12.1? I see just fixes so a patch release should be sufficient

juliusknorr (approved):

PR #1579, opened 2025-06-25
feat: Toggle notes list sidebar

### LINKED ISSUE #1109: Allow hiding of the notes pane in the three-pane UI

I am often on work calls and take meeting notes down on Nextcloud Notes. Sometimes, I might need to share the screen with the note visible, especially if we are discussing changes to a design collateral that I am noting down. I work with different clients, and it is not ideal for clients to be able to see notes taken down for other clients.

With the three pane layout, collapsing the sidebar only collapses the categories pane, and the notes in the middle pane are still visible (if no category is selected, it shows recent notes). In case there are nested categories, and one is selected, then the notes pane will show notes separated by category headers, and in my case this usually means notes for other clients, with the headers being the client names.

I am listing a few solutions we could choose to implement:

1. A toggle in the settings pane allowing you to hide the notes pane in the middle
   
   While I think this would work, it might mean considerable change to the UX and code base for this to happen. The solution proposed by @caasi-dev in #1108 would work on an individual note basis, but I feel it might be better if we could hide the whole pane that shows notes temporarily since this would be an ideal case scenario than hiding notes individually.
   
1. Distraction-free mode
   
   Adding a toggle in the Rich Text Editor's toolbar to "fullscreen" the editing pane for a distraction free mode would also work. This way, for anyone wanting to truly focus on writing something and not be distracted would appreciate this approach.
   
   image  
   Emacs  
   
   The RTE itself can be shown similar to my Emacs Org mode setup pictured above where the content is set it to a specific width. This way, it's easier to read and edit long sets of content, and also have a distraction free approach.  
   
   This would mean, we don't change anything UI wise, the UX for the three-pane layout mostly stays the same, but we would now be able to just show the contents of a particular note over a screenshare without fear of showing sensitive info.
   
1. A new Vue component in the sidebar pane that allows nested navigation of categories similar to details parent and detail screens/pages on mobile apps
   
   We can add a setting in the settings pane giving users the option to switch to a two-pane style layout, where the middle pane is essentially shifted to the first pane, with a top toolbar style navigation for the pane. Here is a quick wireframe I drew up for this; no prototype, but you'll get a fairly good idea of how it would work like: [wireframe](https://design.penpot.app/#/view/a0f6b4c1-6567-8129-8003-1bf60e830ad6?page-id=a0f6b4c1-6567-8129-8003-1bf60e830ad7&section=interactions&index=0&share-id=a0f6b4c1-6567-8129-8003-1bf8ac17ad6f)

From the existing codebase, none.

**Additional context**
Some context provided above in the issue description.

With further discussion on what we can do, I would like to help contribute code. If the issue in #1108 is related, I would also like to help contribute a solution for that as well.

enjeck:
Fixes #1109

### DIFF
--- a/src/components/CategoriesList.vue
+++ b/src/components/CategoriesList.vue
@@ -5,6 +5,9 @@
 
 <template>
 	<Fragment>
+		<NcButton @click="onToggleNotesList">
+			{{ t('notes', 'Toggle Notes List') }}
+		</NcButton>
 		<NcAppNavigationItem
 			:name="t('notes', 'All notes')"
 			:class="{ active: selectedCategory === null }"
@@ -47,6 +50,7 @@ import {
 	NcAppNavigationItem,
 	NcAppNavigationCaption,
 	NcCounterBubble,
+	NcButton,
 } from '@nextcloud/vue'
 import { Fragment } from 'vue-frag'
 
@@ -65,6 +69,7 @@ export default {
 		Fragment,
 		NcAppNavigationItem,
 		NcAppNavigationCaption,
+		NcButton,
 		NcCounterBubble,
 		FolderIcon,
 		FolderOutlineIcon,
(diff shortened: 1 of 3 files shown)


enjeck:
Closing since not acceptable to designers and too many conflicts. Better to redo from scratch

Issue #1587, opened 2025-07-03
🎨 Use outline icons


blizzz:
**Is your feature request related to a problem? Please describe.**

As a Nextcloud user I want Collectives to look modern

**Describe the solution you'd like**

Search and replace icons to their outline counterparts, except app icon and mimetype icons

marcoambrosini:
Hi Julia, if there's no outline alternative we keep them as they are for now

PR #1596, opened 2025-07-18
feat(ui): switch to outline icons

JuliaKirschenheuter:
Fixes https://github.com/nextcloud/notes/issues/1587

For all leftover icons: https://github.com/nextcloud/notes/issues/1587#issuecomment-3089721801


grnd-alt (approved):

marcoambrosini (approved):
Just one small change. As a guideline, we don't want outline version of icons that are made of only lines.

So, the plus icon is made of 2 lines, does not have a fill: 
Screenshot 2025-07-25 at 16 10 09

We should keep that, and not use this:
Screenshot 2025-07-25 at 16 10 19

PR #1614, opened 2025-08-12
fix(performance): Avoid checking existence of default notes folder 3 times

CarlSchwan:
In the case where the user didn't change their default notes folder.

In the worst case, if the user did change their default notes folder we keep the same amount of queries as before.


lib/Service/SettingsService.php
```diff
@@ -41,7 +43,7 @@ public function __construct(
 			'fileSuffix' => $this->getListAttrs('fileSuffix', [...$this->defaultSuffixes, 'custom']),
 			'notesPath' => [
 				'default' => function (string $uid) {
-					return $this->getDefaultNotesPath($uid);
+					return $this->getDefaultNotesNode($uid)['path'];
```
CarlSchwan, line 46: not the biggest fan of that, I wonder if it would make sense to keep the old getDefaultNotesPath as it is for this part

PR #1620, opened 2025-08-21
fix: Check for null setting

enjeck:
We can get the exception: `Trying to access array offset on null at /opt/nextcloud/apps/notes/lib/Service/SettingsService.php#118`

### DIFF
--- a/lib/Service/SettingsService.php
+++ b/lib/Service/SettingsService.php
@@ -115,11 +115,12 @@ class SettingsService {
 			if ($name === 'notesPath' && $value !== null) {
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


enjeck:
too old, will reinvestigate if reported again

Whisprin:
@enjeck I'm getting this warning in my nextcloud log:
```
Trying to access array offset on null at /var/www/html/custom_apps/notes/lib/Service/SettingsService.php#165
Undefined array key "_route" at /var/www/html/custom_apps/notes/lib/Service/SettingsService.php#165
```

I couldn't find an existing issue on this topic. Should I open one?

Issue #1623, opened 2025-08-22
Attachment API questions, possible tweaks


cheywood:
Hey all, great to see the recent merge of the attachments API :+1: 🎉 @oli-ver @enjeck

I'm working to integrate it into [a client](https://apps.gnome.org/en-GB/Iotas/) and have a few questions with some possible user-centric suggestions. My knowledge about attachments in Nextcloud independent of this API is limited though, so I could be misunderstanding :)

**Why is the random string used for API upload filenames, compared to web (`OCA.Text`) using the recognisable name?**

_Problems_ - A readable filename is a better experience for the user in non-WYSIWYG editors. Plus being able to know what the filename is probably going to be removes the need to, after every attachment upload completes, update the text in the note while it's being edited. The server also ends up with a bunch of more usefully named files.

_Suggestion_ - Use the filename when it's unique. Upon name conflicts use the Nextcloud convention if there is one, or browser download conflict style suffixing (as users are familiar with this).

**Why do the API uploads get stored alongside the note, compared to the web client's in a `.attachments.<fileid>` directory?** 

_Problems_ - It lacks consistency with the web UI and arguably creates a bit of a mess. Plus if the discarding of filenames is resolved, having one namespace (per category) is more likely to result in filename conflicts. 

Edit: also, attachments don't get deleted automatically with the note, unlike those in an `.attachments.<fileid>` directory. 

A further edit: the attachments also don't get moved into the category sub-folders, resulting in disconnection from the note (again those with `.attachments.<fileid>` form do).

_Suggestion_ - Do the same thing as the web editor.

**How does one delete an attachment?**

I'm guessing there's no garbage collection as that would mean parsing each markdown file to determine which files are no longer used? Access in the API would be useful.

----

If my suggestions make sense and nobody else is ready to jump on them I can have a look.

I'll hold on integrating changes into the client for now.

Comments are for MR #1600, original issue #826

cheywood:
Hey @enjeck, thanks for getting back to me. Sorry this got a bit long and devil's advocate-y! :)

> The file ID will change if you move your notes to another server. As a consequence, no attachment will work anymore.

Presuming this means Nextcloud has no mechanism for running logic as part of a restoration, providing something like a _Remap Attachments After Migration_ function via the preferences would be one way to cater for this. It can determine each note's old file id from the directory names in the markdown and rename the attachments directory and update the contents.

Does anything exist already for the attachments created by the web and Android?

[...]

Issue #1470, opened 2025-01-27
Attachments in shared notes are deleted after adding a category


mserwan:
**Bug Description**
When a shared note is categorized by the recipient (without changing ownership), all attachments (e.g., images, videos) within the note are deleted from the server. This issue occurs specifically when the shared user modifies the note's category.

**Steps to reproduce**
User A creates a note with attachments (e.g., images, videos) in the Notes app.
User A shares the note with User B.
User B changes the category of the shared note (e.g., assigns it to "Work" or "Personal").
Check the note content after the category change.

**Expected behaviour**
The note and its attachments (images, videos, etc.) should remain intact when the recipient changes the category of a shared note.

**Actual behaviour**
Attachments within the note are deleted from the server, and the note no longer displays them. A "Failed to load" error appears for the missing attachments.

mserwan:
is anyone have idea for this ?

enjeck:
> Whenever a note is shared to someone and the person edits the category, attachments get broken for whatever reason

Likely because the target folder for attachments is got using: `$this->noteUtil->getCategoryFolder($notesFolder, $note->getCategory())`. So attachments are stored in the same category folder as the note (same directory).

PR #1633, opened 2025-09-08
feat: improve attachment storage for API

cheywood:
- Attachments are stored in note-specific folder
- Attachment filenames are retained

This mirrors the behaviour for attachments created in the web UI, bringing the following benefits:
- Attachments get deleted (edit: and moved) along with the note
- Attachments don't clutter the directory alongside the note
- Non WYSIWIG editors which consume the API provide a better experience (via filenames not being replaced by random strings)
- Consistency with web

Most of the code is based on what's used in [Text](https://github.com/nextcloud/text/).

If the API wasn't already public I would switch the name of the return element for image creation to `path` (from `filename`). As there likely isn't any project consuming the API yet this may not be a bad idea.

Addresses part of #1623.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -200,7 +200,7 @@ return ['routes' => [
 		'url' => '/api/{apiVersion}/attachment/{noteid}',
 		'verb' => 'GET',
 		'requirements' => [
-			'apiVersion' => '(v1.4)',
+			'apiVersion' => '(v1|v1.4)',
 			'noteid' => '\d+'
 		],
 	],
@@ -209,7 +209,7 @@ return ['routes' => [
 		'url' => '/api/{apiVersion}/attachment/{noteid}',
 		'verb' => 'POST',
 		'requirements' => [
-			'apiVersion' => '(v1.4)',
+			'apiVersion' => '(v1|v1.4)',
 			'noteid' => '\d+'
 		],
 	],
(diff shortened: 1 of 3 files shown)


enjeck (commented):
Thanks for working on this and for raising question at https://github.com/nextcloud/notes/issues/1623. I need time to read through the previous discussions when the current API was implemented before I can properly review this

enjeck:
> I picked up that this impacts moving notes between categories too; without this fix the attachments are left in the previous location, disconnected from the note.

Yes, also previously reported at https://github.com/nextcloud/notes/issues/1470

enjeck (commented):
Like I mentioned at https://github.com/nextcloud/notes/issues/1623#issuecomment-3471583175, I like the idea of using the original file name instead of a random text. But concerned about the new link structure. Apart from the concerns with using the file ID, I wonder how backwards compatible it is. I wonder if it's easier to just keep the existing struture in the meantime and instead explore moving things around when the category is changed.

cheywood:
> Like I mentioned at [#1623 (comment)](https://github.com/nextcloud/notes/issues/1623#issuecomment-3471583175), I like the idea of using the original file name instead of a random text. But concerned about the new link structure. Apart from the concerns with using the file ID, I wonder how backwards compatible it is. I wonder if it's easier to just keep the existing struture in the meantime and instead explore moving things around when the category is changed.

Responded over at #1623

enjeck:
Thanks for the fix and discussion here and in #1623. I'm happy with this improvement, although it will intentional breaking change for API 1.4

I can't push to this branch, so I opened https://github.com/nextcloud/notes/pull/1998 based on this:

While reviewing, I found a few bugs and fixed them:

- The filename conflict counter started at "(2)" instead of "(1)".
- The extension regex could break on unusual filenames.
- Filenames were checked for validity too late, after the code already touched the filesystem with the raw name.
- Invalid filenames returned a 500 error instead of a 400.
- I also found that moving a note to a new category, or deleting a note, did not move or delete its attachment folder. Both are fixed now, so attachments actually follow the note as intended.
- Added deletion endpoint.

PR #1730, opened 2026-01-26
Use filled favorite icons and remove document icons

aronovgj:
The left‑side FileDocumentOutlineIcon is always shown but doesn't have any function, and it disappears for favorites. This is inconsistent and wastes space. This change turns that area into a visible, clickable favorite toggle, so the star state is always visible and the icon slot has a purpose.

### Changes

* Add a persistent favorite toggle button to the left of each note
* Use outline star for non‑favorites, filled star for favorites
* Remove the favorite action from the dropdown menu

### Why
Keeps the UI consistent (favorites vs non‑favorites), makes the icon slot functional, and improves discoverability of the favorite action.

This is how it would look now:

image

### DIFF
--- a/src/components/NoteItem.vue
+++ b/src/components/NoteItem.vue
@@ -23,14 +23,10 @@
 				:size="20"
 				fill-color="#E9322D"
 			/>
-			<StarOutlineIcon v-else-if="note.favorite"
+			<StarIcon v-else-if="note.favorite"
 				slot="icon"
 				:size="20"
 				fill-color="#FC0"
-			/>
-			<FileDocumentOutlineIcon v-else
-				slot="icon"
-				:size="20"
 			/>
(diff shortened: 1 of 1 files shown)


aronovgj:
Besides, I feel like the favorite button might be justifiably prominent. If you for example look at Google Chrome search bar: the Browser contains loads of features but they decided that the favorite icon specifically should be in the search bar.

marcoambrosini:
> Then you still have an empty yellow star if a note is a favorite. As a user my intuition is that the star is clickable

I would change it back to filled for now, so that it's consistent with how we treat this icon in other places in Nextcloud. I think it got mistakenly changed to outline recently.

marcoambrosini (approved):
Nice, thanks @aronovgj!

PR #1743, opened 2026-02-05
perf: Replace getById call with getFistNodeByid

CarlSchwan:
We only need the first node anyway and getFirstNodeById is quite a bit faster in many cases.


artonge (approved):
Is it also an issue on user folders?

In any case, it simplifies the code, so let's go!

enjeck:
> getFirstNodeById is not available with NC28 :/ I wonder why this app still support 28, it's EOL since a while

i don't mind updating for the next release

enjeck (approved):

PR #1745, opened 2026-02-07
fix: Migrate to new files API

enjeck:


enjeck:
opened https://github.com/nextcloud/notes/pull/1827 instead

PR #1813, opened 2026-04-08
Update PHP max version to 8.5 in info.xml

### LINKED ISSUE #1764: PHP 8.5 currently not supported
I currently can't install the Notes app via the app browser. It only offers a very outdated version (3.5.1).
I also tried installing the app manually, but that failed as well. When I tried to enable the app, the app browser showed: "PHP with a version lower than 8.4 is required".

My Nextcloud instance (33.0.0) is already running on PHP 8.5.

### LINKED ISSUE #1812: Notes on latest nextcloud 33.0.2 (winter 26) broken
If I try to install notes, after a update to nextcloud 33.0.2, I can't enable notes again. It crashes with an exception about not available function.

I checked it again what exactly the problem is and I realize that my nextcloud installation is installing a very outdated version 3.5.1.

Image

the latest version should be 4.13.1, but only version 3.5.1 is available.

Image

I also confirmed that the latest nextcloud version is installed.

HolgerHees:
As nextcloud 33 support php 8.5, notes should support it too.

Current situation is, if I run nextcloud 33 with php 8.5, the latest notes version is 3.5.1, which is totally wrong.

The reason is that version 3.5.1 has a '*' as a php version string. Thats interpreted as version 3.5.1 is supporting all php versions. But version 3.5.1 is not running with nextcloud 33.

So the solution is to change the supported php version to match the one in latest nextcloud.


enjeck (commented):
to resolve the failing CI https://github.com/nextcloud/notes/actions/runs/24134072818/job/70417547051?pr=1813, can you ammend your commit message? Maybe prefix it with "fix: " or "chore: "

HolgerHees:
> to resolve the failing CI https://github.com/nextcloud/notes/actions/runs/24134072818/job/70417547051?pr=1813, can you ammend your commit message? Maybe prefix it with "fix: " or "chore: "

done

joshtrichards (approved):
Separately I wonder if we really need to keep v28 / 8.0 supported here, but don't let that hold this up.

enjeck:
> Separately I wonder if we really need to keep v28 / 8.0 supported here, but don't let that hold this up.

updated to 30/8.1 in another PR. Ideally should be on 32+, but don't wanna make such a big jump suddenly

enjeck (approved):

PR #1841, opened 2026-05-02
fix dead zones in clickable note area

aronovgj:
The upper and lower borders of a note are not clickable although mouse cursor shows a hand

- Make the full visible note row clickable, including the top and bottom edges
- Move the row padding onto the actual anchor so the cursor and click target match


enjeck (approved):
thanks

Issue #1803, opened 2026-03-29
Cannot share a note in Nextcloud webui


MichaelTorfs:
Please make a clear and concise description of what the bug is.

## Steps to reproduce
1. go to the notes web ui
2. click "+ New note"
3. enter some text
4. click 3 dots at the right side of the note
5. click share from the menu that appeared
6. nothing happens

### Expected behaviour
a share screen should appear were users can be selected

### Actual behaviour
nothing happens

### Screenshots

<img width="880" height="496" alt="Image" src="https://github.com/user-attachments/assets/ce729568-85df-4722-b6fb-a8bad25909ae" />

## Server
Please complete the following information.

- Notes app version:  4.13.1
- Nextcloud version: 33.0.1
- OS: Ubuntus Server with Docker (Linux 6.8.0-106-generic x86_64)
- Web server: Apache
- PHP version: 8.4.19
- Database: PostgreSQL 18.3 on x86_64-pc-linux-musl, compiled by gcc (Alpine 15.2.0) 15.2.0, 64-bit

**Nextcloud configuration:**

<details>
</details>

## Client
Please complete the following information.

- Browser (incl. version): Brave 1.88.136
 - OS: Windows 11

## Log files
Around the time I try to click Share the follwing appears in the logs (I'm not sure it's related):

<img width="1811" height="940" alt="Image" src="https://github.com/user-attachments/assets/6156bcdc-a342-43c3-98a7-baf9c0732104" />

guilegatan:
Same here. I see the share-link, but clicking it does nothing.

Notes app version: 4.13.1
Nextcloud version: 33.0.2
OS: Ubuntus Server 24.04.4
Web server: Apache
PHP version: 8.3
Database: MySQL 8.0.45

GuillaumeEvrard:
Same here, very frustrating
Web server: Apache
Notes app version: 4.13.1
Nextcloud version: 33.0.2

enjeck:
fixing at https://github.com/nextcloud/notes/pull/1827

andrewhotlab:
I just upgraded to Notes from version 4.13.1 to 5.0.0, and sharing seems not to be working (obviously I cleared the browser cache, and tried with different browsers and user accounts):

<img width="1101" height="277" alt="Image" src="https://github.com/user-attachments/assets/3d249655-cd4a-4f38-9cee-72807c04ed55" />

[...]

PR #1852, opened 2026-05-09
fix: Update min server version to 33

enjeck:
Because the nextcloud-files library used at https://github.com/nextcloud/notes/pull/1827 requires NC 33+. We'll release v6.0 after this. 

See https://github.com/nextcloud/notes/issues/1803


blizzz (approved):

blizzz:
@enjeck since server master is 35, the tests needs to be adjusted, or compatibility bumped uo to 35 (would recommend that)

PR #1886, opened 2026-06-04
perf(dashboard): avoid full note reads when building excerpts

joshtrichards:
## Summary

Optimize note excerpt generation by reading only a small prefix of the file instead of loading the full note content.

## Changes

- add a lightweight best-effort excerpt source reader by switching from the File API's getContent to own fopen/fread
- read only a bounded prefix sized for excerpt generation
- trim incomplete trailing UTF-8 bytes with `mb_strcut()`
- preserve existing excerpt formatting behavior after markdown stripping
- return an empty excerpt if the lightweight preview read fails, instead of falling back to a full file read
- additional optimization: use pre-compiled string literal for BOM stripping eliminating need for `pack()` at runtime
- also fixes a small bug: `$title` is typed as string and `empty()` ends up being true on things like `"0"`.

## Why

Excerpt generation only needs roughly the first 100 visible characters, so reading the entire note is unnecessary work. This change reduces I/O and memory usage for note list rendering, especially for larger notes or slower storage backends.

## Notes

- this is intentionally a performance-oriented heuristic, not a full-fidelity content read
- markdown-heavy prefixes may occasionally produce shorter previews than before
- full note reads remain unchanged via `getContent()`

### DIFF
--- a/lib/Service/Note.php
+++ b/lib/Service/Note.php
@@ -56,20 +56,26 @@ class Note {
 			);
 			$content = mb_convert_encoding($content, 'UTF-8');
 		}
-		$content = str_replace([ pack('H*', 'FEFF'), pack('H*', 'FFEF'), pack('H*', 'EFBBBF') ], '', $content);
+
+		// Strip Byte Order Marks (BOM) for UTF-8, UTF-16 BE, and UTF-16 LE
+		$content = str_replace(["\xEF\xBB\xBF", "\xFE\xFF", "\xFF\xFE"], '', $content);
+
 		return $content;
 	}
 
 	public function getExcerpt(int $maxlen = 100) : string {
-		$excerpt = trim($this->noteUtil->stripMarkdown($this->getExcerptSource($maxlen)));
+		$excerpt = $this->noteUtil->stripMarkdown($this->getExcerptSource($maxlen));
+
 		$title = $this->getTitle();
 		if ($title !== '') {
-			$length = mb_strlen($title, 'utf-8');
-			if (strncasecmp($excerpt, $title, $length) === 0) {
-				$excerpt = mb_substr($excerpt, $length, null, 'utf-8');
+			$titleLength = mb_strlen($title, 'utf-8');
+			if (strncasecmp($excerpt, $title, $titleLength) === 0) {
+				$excerpt = mb_substr($excerpt, $titleLength, null, 'utf-8');
 			}
 		}
+
 		$excerpt = trim($excerpt);
+
 		if (mb_strlen($excerpt, 'utf-8') > $maxlen) {
 			$excerpt = mb_substr($excerpt, 0, $maxlen, 'utf-8') . '…';
 		}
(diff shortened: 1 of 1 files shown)


joshtrichards:
Lint php-cs failures are unrelated - see #1905

lib/Service/Note.php
```diff
@@ -56,26 +56,64 @@ public function getContent() : string {
...
+		// Strip Byte Order Marks (BOM) for UTF-8, UTF-16 BE, and UTF-16 LE
+		$content = str_replace(["\xEF\xBB\xBF", "\xFE\xFF", "\xFF\xFE"], '', $content);
+
 		return $content;
 	}
 
 	public function getExcerpt(int $maxlen = 100) : string {
-		$excerpt = trim($this->noteUtil->stripMarkdown($this->getContent()));
+		$excerpt = $this->noteUtil->stripMarkdown($this->getExcerptContent($maxlen));
+
```
enjeck, line 68: since this no longer goes through `getContent()`,  it loses the non-UTF-8 handling done there. A UTF-16-encoded note that produced a readable excerpt before will now be read as raw bytes and decoded as UTF-8 below, yielding garbage?

lib/Service/Note.php
```diff
@@ -56,26 +56,64 @@ public function getContent() : string {
...
+	private function getExcerptContent(int $maxlen) : string {
+		$handle = $this->file->fopen('r');
+		if (!is_resource($handle)) {
+			return '';
+		}
+
+		// Over-read bytes assuming worst-case UTF-8 size (up to 4 bytes per
+		// character). This is only a heuristic for preview generation; markdown
+		// stripping may reduce the visible character count further.
+		$bytesToRead = max(512, $maxlen * 4);
```
enjeck, line 97: Maybe  * 6 is better. With the default `maxlen=100` this reads 512 bytes. After `stripMarkdown()` and the leading-title strip (lines 71–76), a long first line / URL / long title can push the visible excerpt below maxlen, where the old full-read produced a complete one

lib/Service/Note.php
```diff
@@ -56,26 +56,64 @@ public function getContent() : string {
...
+			$content = fread($handle, $bytesToRead);
+			if ($content === false) {
+				return '';
+			}
+		} finally {
+			fclose($handle);
+		}
+
+		// Remove any partial trailing multibyte character from the truncated read.
+		$content = mb_strcut($content, 0, strlen($content), 'UTF-8');
```
enjeck, line 109: I dont understand. The comment says this removes a partial trailing multibyte char, but passing strlen($content) (the full byte length) as the cut length makes it effectively a no-op for that purpos?

lib/Service/Note.php
```diff
@@ -56,26 +56,64 @@ public function getContent() : string {
...
+
+		// Strip Byte Order Marks (BOM) for UTF-8, UTF-16 BE, and UTF-16 LE
+		$content = str_replace(["\xEF\xBB\xBF", "\xFE\xFF", "\xFF\xFE"], '', $content);
```
enjeck, line 112: The UTF-16 BOMs (\xFE\xFF, \xFF\xFE) are stripped here, but since the body isn't transcoded from UTF-16  (a i said at https://github.com/nextcloud/notes/pull/1886/changes#r3450190129) , the rest of a UTF-16 note is still mis-decoded?

enjeck (commented):
some comments:

PR #1927, opened 2026-07-06
fix(categories): keep notes when deleting a category

MiMoHo:
## Summary

Deleting a category currently deletes the category folder **including all note files inside it** — a single confirmation click can destroy many notes. This change makes "Delete category" keep the notes: they are moved to *Uncategorized* instead, and only the empty folder goes away.

## How it works

- All notes of the category (including subcategories) are moved to the notes root via the existing per-note category mechanism (`Note::setCategory('')`), so file name collisions are resolved the same way as when changing a single note's category (`Title (2).md`).
- Afterwards, only folders that contain **no files at all** are removed (new helper `deleteFoldersWithoutFiles`). Non-note files that happen to live in the category folder are never deleted.
- The confirmation dialog and the client-side store update reflect the new behavior.

## Note on the behavior change

This intentionally changes the semantics of an existing action, based on the expectation that "deleting a folder" should not silently destroy its notes. If you would rather keep the destructive delete and offer a choice (e.g. "delete notes" vs. "keep notes"), I am happy to rework this accordingly.

## Testing

- Verified end-to-end on Nextcloud 35 (dev): category folder removed, all notes kept, a colliding file name was renamed to `Name (2).md`, non-note files untouched.
- `php -l` clean, production webpack build clean.

---

## 🤖 AI (if applicable)

- [x] This pull request was prepared with assistance from Claude Code (Claude Fable 5 xhigh).
I reviewed and tested the changes myself.

### DIFF
--- a/lib/Service/NotesService.php
+++ b/lib/Service/NotesService.php
@@ -215,15 +215,35 @@ class NotesService {
 			];
 		}
 
-		$parent = $folder->getParent();
-		$folder->delete();
-		if ($parent instanceof Folder) {
-			$this->noteUtil->deleteEmptyFolder($parent, $notesFolder);
+		foreach ($this->getAll($userId)['notes'] as $note) {
+			$noteCategory = $note->getCategory();
+			if ($noteCategory === $category || str_starts_with($noteCategory, $category . '/')) {
+				$note->setCategory('');
+			}
+		}
+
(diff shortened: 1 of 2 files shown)


enjeck (commented):
please attach a video showing this working

enjeck:
Please reopen if ready to show screenshots/videos of this working

PR #1930, opened 2026-07-06
feat(editor): frame the plain editor and reduce side margins

MiMoHo:
## Summary

**Design proposal** for the plain-text editor (EasyMDE): the writing area is currently capped at 47em, left-aligned and floats without any visual boundary, which makes it feel like writing into empty space, with a disproportionately wide unused margin on larger screens.

This change:

- adds a subtle border around the writing field (`--color-border`, follows the light and dark theme),
- lets the field use the available pane width up to 90em (centered), so side margins shrink while line length stays bounded on very wide screens,
- gives the header zone above the field more breathing room,
- removes the frame in distraction-free (full screen) mode, where it would get in the way.

Since spacing and sizing are a matter of taste, I would appreciate a design review — screenshots of the result in light and dark mode are attached:

shot-overview
shot-chromium-category

## Testing

Verified in Chromium and WebKit, light and dark theme; production webpack build clean.

---

## 🤖 AI (if applicable)

- [x] This pull request was prepared with assistance from Claude Code (Claude Fable 5).
I reviewed and tested the changes myself.


enjeck (commented):
@nextcloud/designers Do you approve?

kra-mo (changes requested):
From the design side, I find the border unnecessary. If margins are unexpectedly large, they should just be made smaller/eliminated, no?

jancborchardt (changes requested):
I agree with @kra-mo, the border / framing is unnecessary, it’s not really an element we ourselves use (one exception is Mail where there can be multiple messages in a thread), nor others like Notion use it.

PR #1935, opened 2026-07-15
fix(Helper): do not expose internal exception

silverkszlo:
Refers to: https://github.com/nextcloud/notes-android/issues/3131#issuecomment-4967912348

Do not expose internal exception types

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Controller/Helper.php
+++ b/lib/Controller/Helper.php
@@ -123,9 +123,9 @@ class Helper {
 	}
 
 	/** @param 200|201|400|403|404|423|500|507 $statusCode */
-	public function createErrorResponse(\Throwable $e, int $statusCode) : JSONResponse {
+	public function createErrorResponse(\Throwable $e, int $statusCode, bool $exposeType = true) : JSONResponse {
 		$response = [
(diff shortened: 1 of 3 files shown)


lib/Controller/Helper.php
```diff
@@ -123,9 +123,9 @@ public function logException(\Throwable $e) : void {
 	}
 
 	/** @param 200|201|400|403|404|423|500|507 $statusCode */
-	public function createErrorResponse(\Throwable $e, int $statusCode) : JSONResponse {
+	public function createErrorResponse(\Throwable $e, int $statusCode, bool $exposeType = true) : JSONResponse {
 		$response = [
-			'errorType' => get_class($e)
+			'errorType' => $exposeType ? get_class($e) : 'Exception'
```
provokateurin: You should never expose the type.

provokateurin (changes requested):

lib/Controller/Helper.php
silverkszlo: ok, thanks!

provokateurin (approved):

PR #1933, opened 2026-07-13
(old) Chore/upgrade vue 3

silverkszlo:
🚧 work in progress 🚧

Upgrade to Vue 3

- [x] upgrade dependencies
- [x] replace vuex with pinia
- [x] upgrade to eslint config file
- [ ] tbc
...

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

PR #1953, opened 2026-07-23
Upgrade from Vue 2 to Vue 3

silverkszlo:
- Upgrade Vue 3 and according dependencies (from Pinia 2 to 3, remove v-frag, etc.)
- upgrade eslint config file
- cherry-pick some commits from first draft: https://github.com/nextcloud/notes/pull/1933
- run linter and add manual linting fixes

### 🤖 AI

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,42 +1,45 @@
 {
 	"name": "notes",
+	"version": "6.0.1",
+	"private": true,
 	"homepage": "https://github.com/nextcloud/notes",
 	"license": "agpl",
-	"private": true,
 	"scripts": {
 		"build": "webpack --node-env production --progress",
 		"dev": "webpack --node-env development --progress",
-		"watch": "webpack --node-env development --progress --watch",
-		"lint": "ESLINT_USE_FLAT_CONFIG=false eslint --ext .js,.vue,.ts,.mjs src playwright",
-		"lint:fix": "ESLINT_USE_FLAT_CONFIG=false eslint --ext .js,.vue,.ts,.mjs src playwright --fix",
+		"lint": "eslint src playwright",
+		"lint:fix": "eslint src playwright --fix",
 		"start:nextcloud": "node playwright/start-nextcloud-server.mjs",
-		"test:e2e": "playwright test",
-		"test:e2e:ui": "playwright test --ui",
-		"test:e2e:report": "playwright show-report",
 		"stylelint": "stylelint 'src/**/*.vue' 'css/**/*.css'",
-		"stylelint:fix": "stylelint 'src/**/*.vue' 'css/**/*.css' --fix"
+		"stylelint:fix": "stylelint 'src/**/*.vue' 'css/**/*.css' --fix",
+		"test:e2e": "playwright test",
+		"test:e2e:report": "playwright show-report",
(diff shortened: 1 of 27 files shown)


mejo- (approved):
I reviewed all comments and they look good to me. Really nice work ❤️

One nitpicking comment: "chore(pinia): pinianize app, notes and sync stores" seems to be an empty commit and can probably be dropped?

package.json
```diff
@@ -1,42 +1,45 @@
...
 		"markdown-it": "^14.3.0",
 		"markdown-it-bidi": "^0.2.0",
 		"markdown-it-task-checkbox": "^1.0.6",
-		"pinia": "^2.3.1",
-		"vue": "^2.7.16",
-		"vue-frag": "^1.4.3",
+		"pinia": "^3.0.4",
+		"vue": "^3.5.39",
 		"vue-material-design-icons": "^5.3.1",
 		"vue-observe-visibility": "^1.0.0",
```
enjeck: this only works for Vue 2

src/components/NotePlain.vue
```diff
@@ -13,21 +13,21 @@
...
+						<SyncAlertIcon :size="30" fillColor="var(--color-error)" />
 						{{ t('notes', 'The note has been changed in another session. Please choose which version should be saved.') }}
 					</div>
 					<div class="conflict-solutions">
 						<ConflictSolution
 							:content="note.conflict.content"
 							:reference="note.reference.content"
 							:button="t('notes', 'Use version from server')"
-							@on-choose-solution="onUseRemoteVersion"
+							@onChooseSolution="onUseRemoteVersion"
```
enjeck: lets not prefix events with `on`. I think internally, vue 3 converts this to `onOnChooseSolution`

src/App.vue
```diff
@@ -5,21 +5,21 @@
...
 		<NcAppNavigation :class="{loading: loading.notes, 'icon-error': error}">
 			<template #list>
 				<NcAppNavigationNew
 					v-show="!loading.notes && !error"
 					:text="t('notes', 'New category')"
 					@click="onNewCategory"
-					@dragover.native="onNewCategoryDragOver"
-					@drop.native="onNewCategoryDrop"
+					@dragover="onNewCategoryDragOver"
+					@drop="onNewCategoryDrop"
```
enjeck, line 16: Did you check that drag and drop still works after this change? 🤔

silverkszlo: yes it works

package.json
silverkszlo: nice catch, thanks!

src/components/NotesView.vue
```diff
@@ -124,6 +119,7 @@ export default {
 			showNote: true,
 			searchText: '',
 			creatingNote: false,
+			endOfNotesObserver: null,
```
max-nextcloud: My understanding is that data is handled by the vue reactivity system. We don't need reactivity for the endOfNotesObserver. So no need to add it to data.

In some places in text we use `this.$something` to indicate `$something` is not reactive.

max-nextcloud (approved):
Just checked the two last commits as they had not been reviewed.

Approving with one minor comment

PR #1966, opened 2026-08-06
test: add server-free unit tests for note, path and cursor logic

karlitschek:
The suite so far has been integration-only: everything under tests/api/ boots a real Nextcloud, so there was no way to exercise the app's pure logic without a server, and no test at all covered the functions that turn user input into file names.

composer.json already declared a `test:unit` script pointing at tests/unit/phpunit.xml, but neither the config nor PHPUnit itself was present. This makes that script real:

* tests/unit/phpunit.xml + bootstrap.php — no server, no database, no web server. `composer test:unit` works on a bare checkout. The bootstrap declares `OC\Hooks\Emitter`, which `OCP\Files\IRootFolder` extends but nextcloud/ocp does not ship, the same way tests/stubs/ocp.php already fills gaps for Psalm.
* phpunit/phpunit and doctrine/dbal as dev dependencies. DBAL is needed because mocking `OCP\IDBConnection` reflects over `IQueryBuilder`, whose signatures reference Doctrine's types; the server provides it at runtime.
* `OCA\Notes\` is mapped in autoload-dev — the app relies on Nextcloud's own app autoloader, which is absent outside a server.
* A separate phpunit-unit.yml workflow so these run on every pull request in seconds, independently of the server-backed test.yml.

81 tests covering NoteUtil (category-path normalisation including traversal attempts, title derivation, collision-safe file names, markdown stripping), NotesService (which files count as notes, the folder walk including hidden files and attachment folders, titles from content), Note (title, category, excerpt, BOM and object-storage content handling), `Util::retryIfLocked` and ChunkCursor.

## Changes since the first review

Rebased onto current main, which needed two fixes on its own: `NotesService::__construct` has gained a fourth parameter and `gatherNoteFiles()` a `$showHidden` argument since this branch was opened.

Addressing @enjeck's review:

* Trimmed the suite from 121 tests to 81, and dropped the characterization tests that pinned known-bad behaviour in long docblocks. Three of the behaviours they described are real bugs and are fixed below; the other two — a non-positive `maxRetries`, and an empty custom file extension — describe states no caller can reach, so they are simply gone. Also dropped: a duplicate cursor round-trip, tests of trivial getters, and the thinner rows of several data providers.
* Removed the explanatory comments throughout, including the workflow header flagged inline. What is left in tests/unit/ is a single inline comment — object storage returning `false` rather than `''` for an empty file — plus two short docblocks describing what a helper builds.
* Deduplicated the NoteUtil setup that had been copy-pasted across three test classes into tests/unit/NotesTestCase.php.
* Added tests/unit/ to `php_dirs` so `make lint` covers it the way it already covers tests/api/.

One side effect worth knowing about: installing doctrine/dbal lets Psalm resolve Doctrine's classes for the first time, and it immediately flagged a pre-existing deprecated `Table::setPrimaryKey()` call in a migration. That is recorded in tests/psalm-baseline.xml rather than worked around, since the call is what Nextcloud's own migration API expects.

## Two bugs the tests turned up

Both are long-standing, neither is introduced here, and each is fixed in its own commit with a regression test.

**Nested categories were silently dropped.** `gatherNoteFiles()` merged the recursion's category list into the parent's with the `+` array union. Both operands are sequentially-keyed lists, so the union keeps the left-hand value at every occupied index and discards the rest — which subcategory was lost depended on its position, not its depth. Confirmed against a local Nextcloud 36 instance, where the sidebar received 7 categories although 16 exist. The files union beside it is keyed by file id, which is unique across the tree, so no note was ever lost, and the v1 API discards this list entirely, so there is no client-facing change.

**Excerpts mis-stripped multi-byte titles.** `getExcerpt()` passed a character count to `strncasecmp()`, which counts bytes, so content that merely began with the same first character as the title was mistaken for a repeated title and truncated. On the same instance, a note titled 日本語 with the body "日曜日 is Sunday" was excerpted as "is Sunday". The same commit replaces `empty($title)` in `getSafeTitle()`, which made a note whose first line is exactly `0` come out titled "New note".

Separately, `make lint` had one phpcs error left in it — a stray blank line before the closing brace of `SettingsService` — which is now gone too, so the target passes cleanly.

These three commits widen what was a test-only PR. They are small and self-contained, so say the word if you would rather review them separately and I will split them out.

## 🤖 AI (if applicable)


.github/workflows/phpunit-unit.yml
```diff
@@ -0,0 +1,79 @@
+# SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: MIT
+#
+# Unit tests. Deliberately separate from test.yml: these need no Nextcloud
+# server, no database and no web server, so they finish in seconds and give
+# feedback on every pull request. test.yml remains the place where the HTTP API
+# is exercised against a real instance.
```
enjeck, line 7: Comment is unnecessary, can be removed

enjeck (commented):
Haven't looked deeply yet, but a brief glance shows there are too many tests and I don't think they are all useful. Having too many makes CI slower and harder to maintain. I think we should strip this down to a few necessary tests that test core Notes functionality. 

Also, there are too many long comments. I would expect the code to be readable enough to not need so many comments unless where there are "gotchas" to be aware of

enjeck (changes requested):
^^

AndyScherzinger:
@enjeck thanks for the review — both points should be covered now, and the branch is rebased on current main.

**Fewer tests.** The suite is down from 121 tests to 81, in 683 lines instead of 1390. The characterization tests that pinned known-bad behaviour in long docblocks are gone, along with a duplicate cursor round-trip, tests of trivial getters, and the thinner data-provider rows. It runs in about 50ms, so the `unit-tests` job time is almost entirely PHP setup and `composer install`.

**Fewer comments.** The explanatory blocks are gone throughout, including the workflow header you flagged inline. What is left in `tests/unit/` is one inline comment (object storage returning `false` rather than `''` for an empty file) plus two short docblocks describing what a helper builds.

Writing the tests did turn up three genuine bugs, so those are fixed in their own commits with regression tests rather than pinned: nested categories dropped by a `+` array union, excerpts mis-stripping multi-byte titles because a character count was handed to `strncasecmp()`, and a note whose first line is exactly `0` getting titled "New note". The first two I verified against a local instance before and after.

That does widen what started as a test-only PR, so happy to split the fix commits out if you would rather review them separately.

PR #1968, opened 2026-08-06
➕ Move "New category" next to the categories caption

karlitschek:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before|after

The full-width NcAppNavigationNew button sat above everything else in the sidebar and drew attention away from the note list, even though creating a category is a rare action compared to picking one.

It is now an inline action on the "Categories" caption: NcActions renders a single inline action as one icon button, so it appears as a small folder-plus next to the heading, with the action text as its accessible name and tooltip. No separate overflow menu is rendered, because there are no other actions.

Dropping a note on the caption row still starts a new category containing that note, which is what the old button offered. The row rather than the icon is the drop target: it is a much larger area to aim at than a single icon, and the caption gets its own drop highlight since it is not an .app-navigation-entry and cannot reuse the existing .drop-over rule.

Both the click and the drop path now call CategoriesList.startNewCategory() directly instead of round-tripping through the notes:category:new event bus. The subscription stays in place so anything else can still trigger it.

The action is hidden while the note list is in an error state, matching the old button's `v-show="!loading.notes && !error"`; CategoriesList takes a `disabled` prop for that, since the caption itself is still rendered.

The accessible name is unchanged, so the Playwright helpers that locate the button by role and name keep working.

## 🤖 AI (if applicable)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -12,18 +12,7 @@
 	>
 		<NcAppNavigation :class="{loading: loading.notes, 'icon-error': error}">
 			<template #list>
-				<NcAppNavigationNew
-					v-show="!loading.notes && !error"
-					:text="t('notes', 'New category')"
-					@click="onNewCategory"
-					@dragover="onNewCategoryDragOver"
-					@drop="onNewCategoryDrop"
-				>
-					<template #icon>
-						<FolderPlusIcon :size="20" />
-					</template>
-				</NcAppNavigationNew>
-				<CategoriesList :loading="loading.notes" />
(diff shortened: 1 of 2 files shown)


enjeck (commented):
Pls add screenshots of before vs after

jancborchardt (approved):
Looks good design-wise. The extra CSS shouldn’t be needed but needs @enjeck @silverkszlo’s review

AndyScherzinger (approved):

enjeck (approved):
nice

PR #1971, opened 2026-08-06
📚 Note template support

karlitschek:
notes_templates

## 🤖 AI (if applicable)

### DIFF
--- a/playwright/e2e/note-templates.spec.ts
+++ b/playwright/e2e/note-templates.spec.ts
@@ -1,0 +1,107 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import type { Locator, Page, TestInfo } from '@playwright/test'
+
+import { expect, test } from '@playwright/test'
+import { login } from '../support/login.ts'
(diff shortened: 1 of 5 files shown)


max-nextcloud (commented):
@karlitschek Thanks for looking into this.

I had a quick look at the code changes and they seem reasonable.

Could you signoff the commits to pass the DCO check?

Also playwright tests would be great for this to prevent regressions.

max-nextcloud:
Psalm failures are unrelated. Looks like we need to bump some dependencies.

jancborchardt (approved):
Looks great! :)

max-nextcloud (approved):

kizm0:
hello, how can note templates be defined/created? i use version 6.1 and when creating a new note i do not get any picker/option. as addition there is already a great app which uses text templates for nextcloud text [https://apps.nextcloud.com/apps/text_templates](https://apps.nextcloud.com/apps/text_templates) (currently lacks support for markdown)

AndyScherzinger:
@kizm0 you need the text app active (ships with the server) and have templates defined in the template folders (MD or txt). If you create a text document in the files app and it shows templates thank the same should happen in the Notes app. If this is not the case and creation in files does not offer templates to choose, than something is off if the templates. You either have might be need to create a a template folder or rescan for templates (I belive this is a feature and there is a occ command for, but not sure)

Issue #1131, opened 2023-10-13
linking notes like in Obsidian


Orko79:
Hello!
I love Obsidian's “Backlinks” feature, which displays a list of notes on the side of the screen that can be linked or are linked.

When you enter a word and a note exists with the name of the word you entered, Obsidian automatically suggests linking that note to the word. This will convert the word into a Markdown link to the corresponding note. Click on it to jump to the linked note. You can also see in a list which notes are already linked to the note you are currently editing.

This also works the other way around:
If you create a new note, for example with the name “Nextcloud”, you will see a list of all notes that contain the word “Nextcloud”. By clicking on the corresponding entry in the list you can link the respective word to the note you have just created.

In addition, a graphical view could automatically be generated from these links, showing the relationships between notes as a point cloud. The more notes linked to a single note, the thicker the point. This is a great feature, for example for building a knowledge database or streamlining extensive note collections.

abthediablo:
seems to be similar like #951

Issue #1524, opened 2025-04-15
Can't reference other notes through relative links


blacklight:
## Steps to reproduce

- Create a Markdown file named `index.md` with e.g. the following content

```markdown
# Index

- [Topic A](./A.md)
```

- Create a file named `A.md` in the same directory

- Try and click on the link in `index.md`

### Expected behaviour
The note object linked to `A.md` should be opened

### Actual behaviour
The URL is resolved with respect to the current note ID prefix, so the redirect goes to `/apps/notes/note/A.md`, which doesn't exist.

This is particularly relevant when using something like e.g. [vimwiki](https://github.com/vimwiki/vimwiki) (or anything that allows you to manage Wiki-like notes in Markdown format). Relative URLs in these context are usually resolved to file names, but Nextcloud Notes works with internal note IDs instead.

Would it be possible to add an intermediate resolver that checks if the requested URL points to a file indexed as a note, and in that case redirect the user to the right URL?

blacklight:
@enjeck I know that I can reference notes by ID, but it doesn't answer my use case (I'm just realizing that this should have been a feature request rather than a bug btw).

If I want to access the notes from another client pointing to the same directory (e.g. Obsidian, VimWiki...) then most likely that client knows nothing about Nextcloud note IDs - but it probably knows about filenames.

A good solution that may make things work with both client may look like:

- If a note URL is a relative URL in the format `[0-9]+` then it's most likely a note ID
- If instead it's in the format `some/path/Note.md`, and that path exists, then render the associated note ID

Akslaton:
Aside from typing out a link to another note in the usual markdown format:
I feel that it would make sense to have a "link to note" option under the "link" button in the editor tool bar, or maybe a "link to note" option in the Smart Picker. The Smart Picker and the tool bar can already link and do all kinds of stuff, but it can't create a link to other notes.

eldelacajita:
Using the / smart picker would be fine as long as it allows me to directly search for note titles, instead of having to first search for a "link to note" option and then search again for the title.

If using file names is the most "exportable" link format for other applications, I would use that. If a title is changed, and so is the file name, the ideal solution would be having Notes automatically update all links to that file.

Issue #951, opened 2022-12-22
Implement inter-note-references [[link]]


newhinton:
**Is your feature request related to a problem? Please describe.**
Some editors allow for references to be added to notes. Those references are a link to that specific note.
This allows for fancy things like automatic-mind-map-creation.

**Describe the solution you'd like**
I would like to see notes support double-bracket-references.

**Describe alternatives you've considered**
Not implementing this.

**Additional context**
Obsidian.md allows this, aswell as zettlr.

xofit:
wether double-bracket or mark-downstyle [linkdescription](https://actual.link): I would be very glad to see this feature.
I've seen a discussion elsewhere that it might be difficult to implement stable links, as the regular link equals to note title. But I think it instead of using the name in the url it should be possible to implement it with the numerical note ID, which is shown when looking at notes in the webbrowser.

Another nice thing for me would be, if any internal or external link would be clickable not only in view-mode, but also in edit-mode (e.g. ctrl-click or shift-click). I usually look at my notes in edit-mode (there's always something to add..) and would like to be able to hassle free follow links I have previously placed in that note.

joelimgu:
Hello, I am really interested in this feature, its the last thing keeping me from using this as my only note taking app. I'm totally new to the project so I was wondering if there is any reason this isn't implemented or if some UI/UX discussion needs to take place.

If the only issue is time from the developers I can try and make a pull request if needed. But I would like to know if that's the only thing missing before putting the time to code it.

benfrancis:
This would be super useful for syncing notes created with Obsidian, which supports internal links https://help.obsidian.md/Linking+notes+and+files/Internal+links#Supported+formats+for+internal+links

This is the biggest missing feature for me.

PR #1973, opened 2026-08-06
feat(notes): link notes to each other with markdown links

karlitschek:
There was no way to link one note to another. Now there is, using ordinary markdown links whose target is the note's route:

    [Shopping list](/index.php/apps/notes/note/42)

No custom syntax, so every editor renders them — the rich editor, the markdown preview, and anything else that reads the file. This is the same URL NoteReferenceProvider already matches, so a link pasted into Talk or a Text document still resolves to a rich preview.

The target is the file id rather than the title, so renaming a note cannot break a link to it. Only the visible label goes stale.

* "Copy link to note" in the note's action menu puts the markdown on the clipboard, label included, so linking is a paste rather than hand-writing a URL. It sits in the note list row menu, which is present in every editor mode.
* Clicking a note link routes inside the app instead of reloading the page. In rich mode that goes through Text's openLinkHandler hook; note that providing a handler replaces Text's default, so the fallback for every other link — open in a new tab — is reimplemented rather than lost. In preview mode the click is delegated from the preview container, because the rendered HTML is replaced wholesale on every edit.
* Modified clicks (ctrl, meta, shift, alt, middle button) are left alone so "open in new tab" keeps working, and links to another origin are never treated as note links.

Stale labels are refreshed on rename (NoteLinkService), with two limits, because this edits notes the user is not looking at:

* Only labels that still match the old title are rewritten. A link written as [my weekly shop](…/note/42) is the author's wording, not a stale copy of the title, and is left alone.
* Only explicit renames trigger it. Titles also change through autotitle, which fires while a new note is being typed, and sweeping the whole collection on each of those would be wasteful and surprising. The two are separate controller paths, so only updateProperty('title') is hooked.

A rename therefore costs one pass over the notes folder. That is the same O(n) content read as search, acceptable for an explicit, infrequent action.

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/lib/Service/NoteLinkService.php
+++ b/lib/Service/NoteLinkService.php
@@ -1,0 +1,113 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Notes\Service;
+
+use OCP\IURLGenerator;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Keeps the visible label of note-to-note links in step with the note's title.
+ *
+ * Notes are linked with ordinary markdown links whose target is the note's id:
+ *
+ *     [Shopping list](/index.php/apps/notes/note/42)
+ *
+ * Because the target is an id, renaming note 42 can never break a link to it —
+ * only the label goes stale. This service refreshes those labels.
+ *
+ * Two deliberate limits, because this edits notes the user is not looking at:
+ *
+ *  - Only labels that still match the *old* title are rewritten. If somebody
+ *    wrote `[my weekly shop](…/note/42)` that is their wording, not a stale
+ *    copy of the title, and it is left alone.
+ *  - Only explicit renames call this. Titles also change through `autotitle`,
+ *    which fires while a new note is being typed, and rewriting the whole
+ *    collection on every one of those would be both wasteful and surprising.
+ */
+class NoteLinkService {
+	public function __construct(
+		private NotesService $notesService,
+		private IURLGenerator $urlGenerator,
+		private LoggerInterface $logger,
+	) {
+	}
+
+	/**
+	 * Rewrite `[$oldTitle](<link to $noteId>)` to use `$newTitle`.
+	 *
+	 * Never throws: a rename must not fail because a link could not be tidied
+	 * up. Notes that cannot be read or written are skipped.
+	 *
+	 * @return int number of notes changed
+	 */
+	public function refreshLinkLabels(string $userId, int $noteId, string $oldTitle, string $newTitle): int {
+		if ($oldTitle === '' || $oldTitle === $newTitle) {
+			return 0;
+		}
+
+		$pattern = $this->linkPattern($noteId, $oldTitle);
+		// ${1} is the target, ${2} any trailing slashes it was written with
+		$replacement = '[' . $this->escapeReplacement($newTitle) . '](${1}${2})';
+		$changed = 0;
+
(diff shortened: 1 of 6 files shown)


enjeck:
Might fix https://github.com/nextcloud/notes/issues/1131 and https://github.com/nextcloud/notes/issues/951 and https://github.com/nextcloud/notes/issues/1524

lib/Service/NoteLinkService.php
```diff
@@ -0,0 +1,113 @@
...
+	 * @return int number of notes changed
+	 */
+	public function refreshLinkLabels(string $userId, int $noteId, string $oldTitle, string $newTitle): int {
+		if ($oldTitle === '' || $oldTitle === $newTitle) {
+			return 0;
+		}
+
+		$pattern = $this->linkPattern($noteId, $oldTitle);
+		// ${1} is the target, ${2} any trailing slashes it was written with
+		$replacement = '[' . $this->escapeReplacement($newTitle) . '](${1}${2})';
```
enjeck, line 57: I'm worried about this replacement. What if I edit a note to a title with containing `]`?

enjeck: escapeReplacement only escapes the preg_replace specials `$` and` \`. It does not escape `]`

enjeck: maybe adding something like the replace `[` in `noteLinkMarkdown` helps

lib/Service/NoteLinkService.php
```diff
@@ -0,0 +1,113 @@
...
+				$targets[] = $relative;
+			}
+		}
+
+		$targets = array_map(
+			static fn (string $target): string => preg_quote($target, '/'),
+			array_values(array_unique($targets)),
+		);
+
+		return '/\[' . preg_quote($oldTitle, '/') . '\]\(\s*(' . implode('|', $targets) . ')(\/*)\s*\)/u';
```
enjeck, line 103: same here. What if i have a note titled `Test]`

lib/Service/NoteLinkService.php
```diff
@@ -0,0 +1,113 @@
...
+		private IURLGenerator $urlGenerator,
+		private LoggerInterface $logger,
+	) {
+	}
+
+	/**
+	 * Rewrite `[$oldTitle](<link to $noteId>)` to use `$newTitle`.
+	 *
+	 * Never throws: a rename must not fail because a link could not be tidied
+	 * up. Notes that cannot be read or written are skipped.
```
enjeck, line 46: Not sure about this. `getAll(...)` and `linkPattern(...)` sit outside the per note try/catch, and they could throw

lib/Service/NoteLinkService.php
```diff
@@ -0,0 +1,113 @@
...
+				continue;
+			}
+
+			try {
+				$content = $note->getContent();
+				$updated = preg_replace($pattern, $replacement, $content);
+				if ($updated === null || $updated === $content) {
+					continue;
+				}
+				$note->setContent($updated);
```
enjeck, line 71: do we need to lock while doing this or nah? why or why not?

src/noteLinks.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * @param {number|string} noteId id of the note to link to
+ * @return {string} URL for a markdown link
+ */
+export function noteLinkUrl(noteId) {
+	return generateUrl('apps/notes/note/{noteId}', { noteId })
+}
+
+/**
+ * A markdown link to a note, ready to paste into another note.
```
enjeck, line 30: Is it really ready to past into another note? Cuz this gives a relative link that is likely not immediately useful 🤔

enjeck (commented):
Thanks, it would be nice to have this as it's a popular feature request!

Some questions:

PR #1984, opened 2026-08-12
fix(categories): hide dot-prefixed folders from category list

### LINKED ISSUE #1866: Unwanted .attachments folders in Categories
### Expected behaviour
No .attachments folders 

### Actual behaviour
Since the last update 5.0.0, the notes app shows in the categories these weird .attachments folders which are empty, they reappear when deleting them an reloading the page. I'm using the latest Nextcloud AIO.

### Screenshots

Image

xhon-pelushi:
## Summary
- Skip dot-prefixed folders (e.g. Text editor `.attachments.<id>`) when gathering note categories.
- Stops hidden attachment folders from showing up as empty categories in the sidebar.

Fixes #1866

## Test plan
- [ ] Create a note with an inline image via Text so an `.attachments.*` folder exists.
- [ ] Confirm the Notes sidebar categories list does not include that folder.
- [ ] Confirm normal user categories still appear.
- [ ] `composer run cs:check` / `composer run psalm` if convenient.

### DIFF
--- a/lib/Service/NotesService.php
+++ b/lib/Service/NotesService.php
@@ -250,6 +250,12 @@ class NotesService {
 		$nodes = $folder->getDirectoryListing();
 		foreach ($nodes as $node) {
 			if ($node->getType() === FileInfo::TYPE_FOLDER && $node instanceof Folder) {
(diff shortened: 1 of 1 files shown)


enjeck (approved):
thanks!

PR #1989, opened 2026-08-17
🧪 Add JS unit test setup, covering Util.js

AndyScherzinger:
The repo had no JS test runner. The only way to exercise frontend logic was Playwright - a Docker container, a login and a browser to check what a helper returns. This adds vitest and puts it to work on `src/Util.js`, which has been there all along, is pure, and had no coverage at all.

### What is in here

* **vitest in jsdom**, with `npm run test` and `npm run test:coverage`, and `coverage/` ignored.
* **`.github/workflows/node-test.yml`**, added verbatim from nextcloud/.github like the sixteen others already in `.github/workflows`. It runs the suite and uploads coverage to Codecov.
* **`src/tests/Util.spec.js`**, 33 cases.

### Why Util.js

It is on main already, so this lands without waiting on a feature branch, and every feature branch can then bring the tests for whatever it adds. `escapeHtml()` in particular is the one piece of that file where a mistake is a security bug, and it had no test.

Covered: copyNote's attribute list and its exclusions, the category label, the new-note route check, and the two drag helpers down to their defensive paths - a browser that refuses a mime type, a transfer that advertises no types, a read-only or unknown note. Util.js ends up fully covered by line and by branch.

<details>
<summary>Implementation notes</summary>

* The config is deliberately smaller than the one in text, circles, deck or files_mindmap: no `@vitejs/plugin-vue`, since no component is tested yet.
* jsdom rather than node, because `escapeHtml()` builds an element and because the app logger other modules pull in reads `window` at import time. A component test would want jsdom anyway.
* The npm script is `test` rather than `test:unit` because the organisation's node-test.yml runs `npm run test --if-present` and `npm run test:coverage --if-present`. `test:e2e` keeps its name, so the two suites stay distinguishable.
* The config file is `vitest.config.js` rather than `.mjs`, so it sits with the repo's other configs and inside that workflow's `**.js` paths filter - as `.mjs`, a config-only change would skip the test job.
* Excluded attributes are asserted with `toStrictEqual`, since `toEqual` cannot tell a property set to `undefined` from one that was never assigned - which is exactly what `copyNote()` does with them.

</details>

### Merge order

#1972 and #1974 each carry a spec for the module they add, and those cannot run until this got merged.

## 🤖 AI (if applicable)

### DIFF
--- a/.github/workflows/node-test.yml
+++ b/.github/workflows/node-test.yml
@@ -1,0 +1,112 @@
+# This workflow is provided via the organization template repository
+#
+# https://github.com/nextcloud/.github
+# https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization
+#
+# SPDX-FileCopyrightText: 2023-2024 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: MIT
+
+name: Node tests
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - master
+      - stable*
+
+permissions:
+  contents: read
+
+concurrency:
+  group: node-tests-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  changes:
+    runs-on: ubuntu-latest-low
+    permissions:
+      contents: read
+      pull-requests: read
+
+    outputs:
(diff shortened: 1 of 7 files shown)


enjeck (approved):
cool

PR #2034, opened 2026-09-18
fix(notes-view): keep the open note visible in nested categories

ppardi:
When a note is selected, the current code's visibility check only compared the category the user selected with the first segment of the note's category (if the category is `Work/Clients`, the check would only compare against `Work`). The current navigation only shows top-level categories (`Work`) so the comparison holds under normal circumstances and the bug doesn't manifest. 

But if the category gets nested due to a rename (if the user renames the `Work` category to `Personal/Work` for example), the note visible in the pane will be hidden because the note's category was moved to `Personal/Work`. But the `rootCategory()` function reduces the note's category to just `Personal`. This no longer matches so the note is seen to be outside its own category.

This fix matches the note against the category the user selected and its descendants. This already is being done by the store when the note list is filtered by the selected category. 

I was able to reproduce this on a live NextCloud 35 instance. I watched the note disappear when I renamed a category and made it nested. 

Also 87 unit tests passed and ESLint reports no new problems. The new cases cover: exact match, descendants, nested exact match (the regression), the `Workshop` vs `Work` prefix trap, null selection, and Uncategorized. 

This change was developed with AI assistance (Claude Code). I reviewed the diff and verified the reproduction and the fix myself on a live instance.


ppardi:
Reopened from a topic branch as #2036 so my fork's main can track upstream. Same commit with no changes.

